PR #2, opened 2015-10-01
Improvements for testing relations

johnsca:
Added a base class for making testing relation stubs easier.

Also fix an issue with charm_dir being required, which made testing hard.

### DIFF
--- a/charms/reactive/test.py
+++ b/charms/reactive/test.py
@@ -1,0 +1,69 @@
+import mock
+from unittest import TestCase
+
+
+class ReactiveTestCase(TestCase):
+    @classmethod
+    def _mock(cls, target, repl=mock.sentinel.DEFAULT):
+        if not hasattr(cls, '_cleanups'):
+            cls._cleanups = []
+        patcher = mock.patch(target, repl)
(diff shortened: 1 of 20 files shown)


charms/reactive/test.py
```diff
@@ -0,0 +1,69 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls._mock('charmhelpers.core.unitdata.kv').side_effect = NotImplementedError('Unmocked access to unitdata')
+        cls._mock('charms.reactive.relations.Conversation.join').return_value = cls
+        cls._mock('charms.reactive.relations.RelationBase.conversation').return_value = cls
+
+    @classmethod
+    def tearDownClass(cls):
+        for cleanup in cls._cleanups:
+            cleanup()
```
bcsaller: https://docs.python.org/2/library/unittest.html#unittest.TestCase.doCleanups 
so shouldn't this use the builting addCleanup?

bcsaller:
LGTM with the noted minor. Once you've looked at that, it can be merged

johnsca:
@bcsaller I'd prefer to use addCleanups for sure, but it's not a classmethod.

PR #3, opened 2015-10-03
Example reactive handler written in Rust

cholcombe973:
I'm not sure if this is right but here's how I interpreted the reactive documents.  I might need to submit another pull request to update the docs.  I found them a bit confusing with relation to creating external handlers.  Is --invoke and --test supposed to take arguments?  I was only guessing that they took the name of the function to call as the argument.

### DIFF
--- a/bin/charms.reactive.rs
+++ b/bin/charms.reactive.rs
@@ -1,0 +1,56 @@
+//! This file contains an example reactive handler written in Rust
+extern crate clap;
+extern crate juju;
+use clap::{Arg, App};
+
+fn hello_world()->Result<(), String>{
+    return Ok(());
+}
+
+fn call_fn(name: String, registry: &Vec<juju::Hook>)->Result<(),String>{
+    for hook in registry {
+        if hook.name == name{
+            return (*hook.callback)();
+        }
+    }
+    return Err(format!("Warning: Unknown callback for hook {}", name));
+}
+
+fn main(){
+    let mut hook_registry: Vec<juju::Hook> = Vec::new();
+        //Register our hooks with the Juju library
+        hook_registry.push(juju::Hook{
+        name: "hello_world".to_string(),
+        callback: Box::new(hello_world),
+    });
+
+    let matches = App::new("myapp")
+        .arg(Arg::with_name("test")
+            .short("t")
+            .long("test")
+            .takes_value(true))
+        .arg(Arg::with_name("invoke")
+            .short("i")
+            .long("invoke")
+            .takes_value(true))
+        .get_matches();
+
+    //Was invoke called?
+    match matches.value_of("invoke"){
+        Some(invoke_arg) => {
+            let result = call_fn(invoke_arg.to_string(), &hook_registry);
+            //handle result
+        },
+        None => {},
+    };
+
+    //Was test called?
+    match matches.value_of("test"){
+        Some(test_arg) => {
+            let result = call_fn(test_arg.to_string(), &hook_registry);
+            //handle result
+        },
+        None => {},
+    };
+
+}


bcsaller:
I don't mean this to be rude, but I don't understand the motivation for this at all. I have only a passing familiarity with Rust but I don't understand to the motivation for walking down this road? 

This doesn't seem to manage the complexity of writing charms better than even the bash examples we have at http://pythonhosted.org/charms.reactive/#non-python-reactive-handlers. If the idea is speed then I'd argue that it doesn't matter in a distributed p2p orchestrated system like Juju where the operations performed on the workload, like restarting the service or downloading new versions of the payload drastically outweigh the time spend doing flow control in the charm. 

I'd like to talk about the motivation, and I'm open to being convinced here, but I am missing the goal looking at thiis.

cholcombe973:
Well the documentation mentions that any language can be used to implement an external handler.  I was just trying to provide an example of that in a statically compiled language.  Currently the examples are only Python or Bash.  I'd be happy to take a stab at writing this in Go or Java or what have you if you'd like that better.

johnsca:
I do think that this is a useful example, because we want to ensure that we can use the reactive framework from other languages, and integrate handlers in those languages with handlers in other layers that are written in different languages.

@cholcombe973 Did you read the documentation of the [ExternalHandler class](http://pythonhosted.org/charms.reactive/charms.reactive.bus.html#charms.reactive.bus.ExternalHandler)?  It is linked to in the section that @bcsaller referenced, and it does describe the interface requirements in more detail.  But perhaps that information should be moved up to (or duplicated in) the front page section.

Specifically, the interface is such that the handler need only respond to --test and --invoke.  Any output from the --test call will be passed through as an argument to the --invoke call and can be used by the handler to dispatch to sub-handlers, like the Bash version does.  No prescriptions are made about what that output should be, as long as the handler can use it to uniquely identify which sub-handlers should be invoked.

However, that's entirely optional and it's also perfectly valid to put one handler per file and not use any arguments to the --invoke flag.  Note, though, that this would cause more subprocess calls and thus will be noticeably slower.

johnsca:
To clarify, `--test` will never be passed an argument, and indicates that the handler should perform checks for each sub-handler (if any) to see which should be invoked.  This is done as a separate step from invoking them so that the reactive framework can avoid reinvoking handlers as it processes new handlers triggered by states set by previous handlers.  During this phase, the handler can use the reactive CLI (which I just realized is completely omitted from the online docs; you can check out [cli.py](https://github.com/juju-solutions/charms.reactive/blob/master/charms/reactive/cli.py) and [bus.py](https://github.com/juju-solutions/charms.reactive/blob/master/charms/reactive/bus.py) to see what's available due to being decorated with `@cmdline.subcommand()`, until I get the docs fixed) to access the state system and perform various checks to determine what should run.  The handler should then exit 1 to indicate that it should not be called again with `--invoke`, or exit 0 with optional output to indicate that it *should* be called again with `--invoke` (and any output passed through).

If called with `--invoke`, the handler need not do any checks, and should just execute the handler function(s).  If it contains multiple handlers, then the output / argument pass-through can be used to track which ones, specifically.

johnsca:
I would like to see the `main` and `call_fn` portions broken out into a crate that could be used to very easily define handlers in Rust, in much the same way that `reactive.sh` encapsulates the dispatch and flag handling logic for Bash handlers.  But I think this is a great start.

galgalesh:
@cholcombe973

Are you still working on this? I'm trying to clean up old PRs

stub42:
We can keep this sort of example in the charms.reactive tree, along with bash. However, we need it integrated with the test suite to ensure the code does not atrophy.

I'm closing this for now, unless there is interest in working on adding this infrastructure.

cholcombe973:
@galgalesh no I haven't touched this code in a long time.  You can close this.

PR #6, opened 2015-10-13
Update README.rst

battlemidget:
Previous link was broken, this points to the only documentation I could find relating to composing a charm with the latest charm.reactive code


johnsca:
The language is changing from "composing" to "building," so this will need to change again soon, but thanks for fixing the link in the meantime.

PR #19, opened 2015-11-06
Issue18

stub42:
Invoke hookenv._run_atstart() after discovery but before dispatch.

I need this to integrate charmhelpers.coordinator with the reactive framework, so that charms can request locks using set_state, and handlers invoked after the leader grants locks using standard @when decorators.


johnsca:
Thanks!

PR #20, opened 2015-11-06
Preflight

stub42:
This branch pulls hookenv.atstart() functionality into the reactive framework. Is it worth it?

A @preflight decorated function can also use @only_once, and probably @when decorators (using state from the previous hook, but I can't think of use cases for mixing with @when). It also gets the same logging and tracing as any other handler.

The downside is of course making charms.reactive more complex. People can always just use hookenv.atstart to register a callback.

Alternative names for the decorator would be setup or startup.

A corresponding @cleanup decorator, run after the main handlers have become quiescent, would also be possible and trivial if we don't allow it to be mixed with @when.


stub42:
I've been tinkering with porting a larger charm, and am finding I need this quite a bit. Code I want to factor out as layers needs to initialize state before hooks are run. Using a decorator with reactive discovery and dispatch seems much easier to test than hookenv.atstart and its global state set on import.

johnsca:
I'm still not clear what kind of "state initialization" you need before hooks are run, or why you couldn't handle it with chained progression of states if you really need something like that?  Is the stuff you're working on in repos that I could take a look at?

stub42:
Here are three components (all untested at this point) that I would want in separate layer(s) to the main charm, as I need exactly the same functionality in both the PostgreSQL and Cassandra charms:

https://git.launchpad.net/~stub/charms/+source/postgresql/tree/reactive/leadership.py?h=reactive
https://git.launchpad.net/~stub/charms/+source/postgresql/tree/reactive/workloadstatus.py?h=reactive
https://git.launchpad.net/~stub/charms/+source/postgresql/tree/reactive/coordinator.py?h=reactive

These are all initializing reactive state from the hook environment, which may not be the same as the previous time a hook ran (or maybe this is the first hook ever run on the unit).

Coordnator fits in really nicely. Using reactive states gives it a much better interface. A handler can just do `coordinator.acquire('foo')`, and then another handler guarded with `@when('coordinator.granted.foo')` will be fired once the leader has granted the lock (in a future hook, unless the current unit happens to be the leader).

Also, https://git.launchpad.net/~stub/charms/+source/postgresql/tree/reactive/apt.py?h=reactive (another potential layer) needs to validate config before any hooks are run, and if it is bad block the unit. The main charm also has a similar validation step for the config it is responsible for. By validating the config in once place once before any hooks or handlers are invoked means that the hooks and handlers don't all need to be stuffed full of error handling. This use case may disappear once `@when` and `@hook` decorators can be mixed, depending on how the scheduling is done.

johnsca:
After getting a better understanding of your intent with leadership (and config) states, I finally see where you are going with this.  As far as naming, our current phases are "hook" and "other"; I think "other" should probably be changed to "states" or "main".  For the new states / decorators, I think I'm partial to `@setup` and `@teardown`.

stub42:
https://git.launchpad.net/~stub/charms/+source/postgresql/tree/lib/preflight.py?h=reactive-wip is simple alternative that could exist outside the reactive framework, perhaps in charmhelpers. It does suffer from Issue #33 though (just a little more work needed to avoid that bug). I lean towards a separate reactive phase though, where the ordering is explicitly laid out setup -> hook -> general and can share behaviour such as logging.

stub42:
ping. I'd like to use `@setup` in the apt and leadership layers instead of the current hack.

stub42:
I'd like to get this MP moving forward (it was included in https://github.com/juju-solutions/charms.reactive/pull/66, so I hadn't mentioned it).

The apt, snap, leadership & coordinator layers all rely on the workaround, as well as charms like PostgreSQL which need to initialize the environment before the main loop is invoked, and I recall others copying the workaround. All charms using these layers or the workaround will stop working if the somewhat fragile workaround starts failing. By moving this into a decorator, we can not only avoid the fragile and  ugly hack, but can keep things working as charms.reactive evolves.

(I've also toyed with the idea of updating the basic layer to discover and execute 'bootstrap' code from layers, which would allow functionality like this to come from a layer rather than charms.reactive. But I think `@setup` is cleaner in charms.reactive, and a common enough need to warrant it. And I worry about unitdata.kv() flushing, so it still might need charms.reactive work)

johnsca:
Closing this in favor of triggers (#121 & #122)

PR #21, opened 2015-11-06
Adjust PATH and PYTHONPATH to work around bundled deps

johnsca:
Because the deps in basic are installed into $CHARM_DIR/{lib,bin} and not at the system level, bash handlers will fail inside `charms.reactive.sh` without this work-around.

### DIFF
--- a/bin/charms.reactive.sh
+++ b/bin/charms.reactive.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+export PATH=$PATH:$CHARM_DIR/bin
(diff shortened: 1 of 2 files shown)


marcoceppi:
LGTM, tested and working

chuckbutler:
:+1:  I witnessed this work

marcoceppi:
:boat:

chuckbutler:
:cow: :boy:

kat-co:
Thank you so much eco team :) :100:

PR #23, opened 2015-11-09
Fix bash handlers not being invoked

johnsca:
Work-around for [lp:1503039 "JUJU_HOOK_NAME does not get set"](https://bugs.launchpad.net/juju-core/+bug/1503039)


marcoceppi:
To be honest this should be the way to do this going forward, regardless of the environment variable as in debug hooks if you attempt to run hooks/install in a config-changed context in 1.25 you'll only get config-changed.

LGTM +1

PR #35, opened 2015-11-19
Set JUJU_HOOK_NAME only when necessary

stub42:
If JUJU_HOOK_NAME is set, we should not override it. This makes 'pdb hooks/install' and similar work. It is also the desirable behaviour for when the cited bug is addressed and Juju always sets JUJU_HOOK_NAME.


johnsca:
Agreed.  Thanks!

PR #37, opened 2015-11-19
Fix #36 - Improve docs based on feedback

### LINKED ISSUE #36: Comments from quickly reading the documentation
Review of http://pythonhosted.org/charms.reactive/

- "The structure of a reactive charm is similar to existing charms, with the addition of reactive and relations directories under hooks:"
This does not agree with the figure below. Only the relations is under hook.

- "The hooks will need to call reactive.main()" all the hooks should do this call once?

- "If using Charm Composition, as is recommended" this needs a reference/link.

- I think we are missing a full example here or at least a pointer to a demo reactive charm/relation.

johnsca:
Address doc feedback in #36

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -40,7 +40,8 @@ Structure of a Reactive Charm
 -----------------------------
 
 The structure of a reactive charm is similar to existing charms, with the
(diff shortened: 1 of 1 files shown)


marcoceppi:
+1 :boat:

PR #39, opened 2015-12-03
source charms.reactive.sh better

kwmonroe:
The example doc for bash handlers assumes charms.reactive.sh is in the CHARM_DIR/bin directory.  Source `which charms.reactive.sh` instead.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -140,7 +140,7 @@ There are helpers for writing handlers in bash.  For example:
 .. code-block:: bash
 
     #!/bin/bash
-    source `which charms.reactive.sh`
+    source charms.reactive.sh
 
     @when 'db.database.available' 'admin-pass'
     function render_config() {
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -140,7 +140,7 @@ There are helpers for writing handlers in bash.  For example:
 .. code-block:: bash
 
     #!/bin/bash
-    source $CHARM_DIR/bin/charms.reactive.sh
+    source `which charms.reactive.sh`
 
     @when 'db.database.available' 'admin-pass'
     function render_config() {


mbruzek:
Hey Kevin,
My understanding of which is it searches the path for the first instance of the match. If charms.reactive.sh is on the path already why use which?

mbruzek:
```bash
root@ip-172-31-22-252:~# which charms.reactive.sh
/usr/local/bin/charms.reactive.sh
root@ip-172-31-22-252:~# which -a charms.reactive.sh
/usr/local/bin/charms.reactive.sh
root@ip-172-31-22-252:~# charms.reactive.sh
This file contains helpers for developing reactive charms in Bash.

Usage:

    source `which reactive.sh`

    @when db.ready cache.ready
    function write_configuration() {
        db_dsn=$(state_relation_call db.ready uri)
        cache_uri=$(state_relation_call cache.ready uri)
        chlp render_template db_dsn=$db_dsn cache_uri=$cache_uri
    }

    @when_not db.ready cache.ready
    function report_blocked() {
        status-set blocked "Waiting for db and/or cache"
    }

    reactive_handler_main

Helpers available:

  @hook PATTERN                Run decorated function if hook matches
  @when STATES                 Run decorated function if states are active
  @when_not STATES             Run decorated function if states not active
  @when_file_changed FILENAME  Run decorated function if file(s) changed
  @only_once                   Run decorated function only once
  set_state                    Set / activate a state
  remove_state                 Remove a state
  relation_call                Call a method on a relation by relation name
  state_relation_call          Call a method on a relation by active state
  all_states                   Check all states are active
  any_states                   Check any states are active
  name_relation_get            Get a relation value by relation name (vs ID)
```

mbruzek:
My point or question is do you even need the which?

johnsca:
@mbruzek I honestly didn't realize "source" would search the path.  +1 for leaving off `which`

mbruzek:
+1 for educating Kevin

PR #40, opened 2015-12-04
Allow filename arguments to be a callable, defering their generation.

stub42:
Allow filename arguments to any_file_changed be callables, so the generation of the filename is deferred until the check is being made.

I currently have a handler decorated with @when_file_changed(postgresql_conf_path()). postgresq_conf_path generates the correct filename using the service configuration. This means the module cannot be imported by my test harness without first installing a mock (which is not a good option, as that mock may not be suitable for all tests). By changing this to @when_file_changed(postgresql_conf_path), my module is easily importable and easier to test.


johnsca:
:shipit: Thanks!

PR #41, opened 2015-12-14
Fix issue regarding duplicate state setting, and adding helpers

johnsca:


charms/reactive/relations.py
```diff
@@ -470,7 +494,8 @@ def set_state(self, state):
             'relation': self.relation_name,
             'conversations': [],
         })
-        value['conversations'].append(self.key)
+        if self.key not in value['conversations']:
```
johnsca: This is the fix.

charms/reactive/relations.py
```diff
@@ -499,6 +524,31 @@ def remove_state(self, state):
...
+        value = get_state(state)
+        if not value:
+            return False
+        return self.key in value['conversations']
+
+    def toggle_state(self, state, active=None):
+        """
+        Toggle the given state for this conversation.
+
+        If ``active`` is not given, the state will be flipped from its current
```
mbruzek: Flipped is a bit confusing here, I would explicitly call out the state will be removed or set.

mbruzek:
This looks like a great fix and includes docstrings!  The "flipped" concept was a bit tough to get until I looked at the code and realized that it was just set_state() or remove_state().  Someone reading the docstring might not understand that.  Just a minor wording problem should not prevent this fix from landing.

+1  from me

PR #43, opened 2015-12-16
Fixed bug in when_not logic

johnsca:
Also added perhaps less confusing aliases: `@when_all` and `@when_none`

### DIFF
--- a/charms/reactive/__init__.py
+++ b/charms/reactive/__init__.py
@@ -26,7 +26,9 @@ from .relations import scopes  # noqa
 from .relations import RelationBase  # noqa
 from .decorators import hook  # noqa
 from .decorators import when  # noqa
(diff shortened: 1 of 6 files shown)


johnsca:
Note: Intended behavior of `@when_not` was to trigger if none of the given states were active.  I.e., this:

```python
@when_not('foo', 'bar', 'qux')
def foo():
    pass
```

Should have behaved the same as:

```python
@when_not('foo')
@when_not('bar')
@when_not('qux')
def foo():
    pass
````

Much like the behavior of `@when` with multiple params vs stacked.  However, the multi-param version was instead triggering if **any** of the give states were not active (even if some of them were).

andrewdmcleod:
looks good

marcoceppi:
LGTM :boat:

Issue #48, opened 2016-01-19
relation hooks without RelationBase


stub42:
Migrating code from pre-reactive means relation hooks without a RelationBase.

A hook definition like this works:

``` python
@hook('{interface:nrpe-external-master}-relation-changed',
      '{interface:local-monitors}-relation-changed')
def enable_nagios():
    [...]
```

But this fails, with an exception about hooks/relations/foo/provides.py not existing:

``` python
@hook('{interface:block-storage}-relation-changed')
def handle_storage():
   [..]
```

I'm not sure why. Hypen in the interface name?

```
Traceback (most recent call last):
  File "hooks/data-relation-changed", line 65, in <module>
    default_hook()
  File "hooks/data-relation-changed", line 56, in default_hook
    main()
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/__init__.py", line 67, in main
    bus.dispatch()
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/bus.py", line 397, in dispatch
    _invoke(hook_handlers)
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/bus.py", line 388, in _invoke
    handler.invoke()
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/bus.py", line 268, in invoke
    args = self._get_args()
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/bus.py", line 262, in _get_args
    return list(chain.from_iterable(self._args))
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/decorators.py", line 58, in arg_gen
    rel = RelationBase.from_name(hookenv.relation_type())
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/relations.py", line 190, in from_name
    relation_class = cls._find_impl(role, interface)
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/relations.py", line 207, in _find_impl
    module = _load_module(filepath)
  File "/var/lib/juju/agents/unit-postgresql-14/charm/lib/pypi/charms/reactive/bus.py", line 442, in _load_module
    sys.modules[modname] = load_source(modname, realpath)
[...]

Issue #33, opened 2015-11-18
Discovery importing dependent modules multiple times


stub42:
If I have reactive/a.py and reactive/b.py that both import lib/shared.py, shared.py ends up being imported twice times and any side effects run twice. Global state will be reset, potentially discarding modifications made by a.py or b.py. atstart/atexit handlers will be registered multiple times.

Work around is to import the modules from the hooks/ stub.

johnsca:
Re-imports are supposed to handled automatically.  I'm guessing this is related to #31 and we need switch to use `import_module` instead of `load_module`; I used the former due to not needing to change PYTHONPATH / sys.path, but I think that it is much less robust and we should just do the path adjustments.

stub42:
I believe the multiple imports are happening because:

```python
modname = realpath.replace('.', '_').replace(os.sep, '_')
sys.modules[modname] = load_source(modname, realpath)
return sys.modules[modname]
```

Here, realpath is the full path to the file, so the module gets imported into sys.modules['var_lib_juju_[...]_reactive_foo.py'] instead of just the module name.

Issue #31, opened 2015-11-17
Discovery breaks relative imports


stub42:
If a file reactive/postgresql/service.py does 'from . import helpers', it fails despite **init**.py files existing in both the reactive and reactive/postgresql directories and the charm directory in sys.path. Absolute imports work fine:

```
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40   File "/var/lib/juju/agents/unit-postgresql-1/charm/hooks/install", line 59, in <module>
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40     default_hook()
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40   File "/var/lib/juju/agents/unit-postgresql-1/charm/hooks/install", line 53, in default_hook
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40     main()
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40   File "/var/lib/juju/agents/unit-postgresql-1/charm/lib/pypi/charms/reactive/__init__.py", line 54, in main
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40     bus.discover()
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40   File "/var/lib/juju/agents/unit-postgresql-1/charm/lib/pypi/charms/reactive/bus.py", line 429, in discover
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40     _register_handlers_from_file(filepath)
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40   File "/var/lib/juju/agents/unit-postgresql-1/charm/lib/pypi/charms/reactive/bus.py", line 448, in _register_handlers_from_file
unit-postgresql-1[4662]: 2015-11-17 09:27:17 INFO unit.postgresql/1.install logger.go:40     _load_module(filepath)
[...]

PR #51, opened 2016-02-03
Update Python3 behaviour to use the standard import mechanism during discovery

stub42:
My charm needs to import modules from other layers. For example, reactive/postgresql/replication.py needs to import reactive/leadership.py. Per Issue #31, relative imports don't work at the moment so 'from .. import leadership' doesn't work. And I can't use absolute imports, because neither CHARM_DIR nor CHARM_DIR/reactive are on the PYTHON_PATH. And there are other glitches with the current discovery implementation, such as Issue #33 and Issue #48 .

This branch replaces the import mechanism under Python3 with one addressing all these concerns. sys.path is adjusted so discovered modules can do imports as they expect (both relative and absolute). The standard import method is used (via importlib.import_module), and the modules imported into either the 'reactive' or 'relations' package namespace. So in my original example, both 'from .. import leadership' and 'from reactive import leadership' work.

I left the old implementation in place for Python2. Is it worth tearing out Python2 support completely, or is it still worth keeping around?


stub42:
Lack of absolute imports is making the apt and leadership layers difficult to use, and is blocking the coordinator layer from being announced. I might be able to work around this if we can't do absolute imports for some reason, by shuffling the API from reactive/leadership.py to lib/reactive/leadership.py (leaving the handlers and bootstrap where it is).

stub42:
Ha. The symlink hack works (cd lib; ln -s ../reactive .), so I have a work around.

stub42:
Shaved hair

stub42:
I've worked out a solid work around, which I'm successfully using in the PostgreSQL charm allowing me to split things into more manageable chunks. I've also moved the bulk of the apt and leadership layers into lib/charms so they no longer need this.

I still think it is worth fixing though, to fix the other import glitches and avoid the work around for charms that wish to import from one reactive/*.py into another.

freeekanayaka:
I've been caught by the same issues.

In absence of regular import mechanism at discover() level, in my case it's even harder to put in place workarounds than with postgres, because I have to import from paths that contain dashes. For example consider the following charm-built layout:

    ./hooks/
    ./hooks/relations/
    ./hooks/relations/jenkins-slave/
    ./hooks/relations/jenkins-slave/requires.py
    ./hooks/relations/jenkins-slave/lib/master.py
   ./hooks/relations/jenkins-slave/unit_tests/test_master.py

In such situation it turns out that:

1. From withing requires.py it's not possible to import utilities living in lib/master.py
2. When discover() scans the tree, it will try to import test_master.py, which fails to import not only because of 1., but also because it contains test-only dependencies that are not declared in wheehouse.txt (since they are not needed at runtime).

freeekanayaka:
Stuart and I discussed the issue with dash-containing interfaces on IRC and agreed that a reasonable solution is to convert dashes to underscores when copying the files to ./reactive, so the tree becomes a regular Python code tree.

Plus, discover() should skip test files (I guess "anything matching test_*.py" is a good enough euristics).

I think this is an important fix, since this issue hinders proper code factoring in reactive-based charms. If nobody is availale to implement the proposed changes in Stuart's branch, please let me know and I'll help.

stub42:
importlib.import_module() is quite happy importing modules and packages with hyphens in the name, so interface names containing hyphens need not block this branch from landing.

Getting discover() to ignore tests and switching from hooks/relations/interface-name to hooks/relations/interface_name can be done in two separate branches (and are probably two separate issues).

freeekanayaka:
Thanks for checking Stuart! Makes all sense. Hope this will be reviewed and merged.

stub42:
Another thing this fixes: it allows pdb to set breakpoints in reactive handlers.

stub42:
More tests demonstrating imports really happening (no mocks), and handling of paths containing hyphens (common in interface names).

simonklb:
I encountered the same issue when trying to do relative imports. Getting this fix in would be great!

Issue #54, opened 2016-02-25
Framework tries to run all files marked as executable and crashes on .md and .yaml files


galgalesh:
Hi

If an interface layer or a reactive directory contains files such as README.md that are marked as executable, then you get the following error:

```
Traceback (most recent call last):
  File "hooks/install", line 19, in <module>
    main()
  File "/usr/local/lib/python3.4/dist-packages/charms/reactive/__init__.py", line 67, in main
    bus.dispatch()
  File "/usr/local/lib/python3.4/dist-packages/charms/reactive/bus.py", line 397, in dispatch
    hook_handlers = _test(Handler.get_handlers())
  File "/usr/local/lib/python3.4/dist-packages/charms/reactive/bus.py", line 381, in _test
    return list(filter(lambda h: h.test(), to_test))
  File "/usr/local/lib/python3.4/dist-packages/charms/reactive/bus.py", line 381, in <lambda>
    return list(filter(lambda h: h.test(), to_test))
  File "/usr/local/lib/python3.4/dist-packages/charms/reactive/bus.py", line 332, in test
    proc = subprocess.Popen([self._filepath, '--test'], stdout=subprocess.PIPE, env=os.environ)
  File "/usr/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.4/subprocess.py", line 1457, in _execute_child
    raise child_exception_type(errno_num, err_msg)
```

I think it's logical that the framework tries to execute all files. However, I think it should show an error message that is a lot clearer. I think this is a problem that a lot of new people will run into. Marking all files as executable is something that new people often do. I don't think we should guard them from mistakes like this but the least we can do is better explain to them what is wrong. If you agree, I can create a patch that returns a clearer error message.

Kind regards
Merlijn

galgalesh:
Maybe we can even throw a warning during build time?

marcoceppi:
I think it's reasonable to also blacklist some extensions, like .md, from
even being considered during evaluation of files to poke. A build-time
warning may also be helpful

galgalesh:
So, we finally found why these files were marked executable in the first place. This happened in a vagrant box. This repository was in a shared folder from a Windows host. Apparently, the permissions on a shared folder like this are interpreted as `a+rwx`. So it is very important that the reactive framework can handle this if we want to support the vagrant boxes...

johnsca:
[...]

PR #55, opened 2016-02-25
Blacklist common non-executable files

galgalesh:
- Blacklist common non-executable files. 
 - Better error message on `OSError [Errno 8]`

I did a quick check on interfaces.juju.solutions to see what non-exec files are present in interface repos. I'm not sure if this is the best place for the no_exec_blacklist tuple since `charm build` is going to need the same list for its warnings. See https://github.com/juju-solutions/charms.reactive/issues/54

### DIFF
--- a/charms/reactive/bus.py
+++ b/charms/reactive/bus.py
@@ -444,6 +444,13 @@ def _load_module(filepath):
 
 
 def _register_handlers_from_file(filepath):
+    no_exec_blacklist = (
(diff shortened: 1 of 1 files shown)


johnsca:
This looks entirely reasonable to me, though I'd also like to fix it to not try to load external handlers from interface layers, which wouldn't work anyway.  I'll probably create a separate PR for that.

PR #56, opened 2016-03-01
Add @when_any and @when_not_all decorators

johnsca:


tests/test_helpers.py
```diff
@@ -222,6 +222,29 @@ def test__when_all(self):
         self.kv.set('reactive.dispatch.phase', 'other')
         assert test(), 'when_all: other; both'
 
+    def test__when_any(self):
```
bcsaller: Is there a reason some of the tests have double under test__ prefixes?

bcsaller:
LGTM, thanks

tests/test_helpers.py
johnsca: The function being tested is `_when_any` but it's not strictly necessary.

chuckbutler:
+1 LGTM -- however I have only run the tests, ymmv

PR #58, opened 2016-03-07
Fix #28: Conversations should be scoped by relation ID

### LINKED ISSUE #28: charms.reactive.relations assumes the relation name is a key
The RelationBase and Conversation model assumes that a relation name name is a key, rather than using the relation id + unit name. This will cause a reactive charm to fail, depending how the related service chooses to use it. For example, here is a metadata.yaml snippet from a client that stores its data in three separate silos:

```yaml
requires:
    session:
        interface: pgsql
    public:
        interface: pgsql
    confidential:
        interface: pgsql
```

On the client side, everything is fine. However, on the server side we have this:

```yaml
provides:
    db:
        interface: pgsql
```

If the client is related three times to the same server, which would be common for development and staging, then the reactive model fails as the relation name can not be used as a key to identify which of the three db interfaces is needed. The relation id must be used.

johnsca:
Fixes #28

### DIFF
--- a/charms/reactive/__init__.py
+++ b/charms/reactive/__init__.py
@@ -36,6 +36,7 @@ from .decorators import only_once  # noqa
 from .decorators import when_file_changed  # noqa
 
 from . import bus
+from . import relations
 from charmhelpers.core import hookenv
 from charmhelpers.core import unitdata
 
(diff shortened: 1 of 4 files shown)


bcsaller:
LGTM, thanks

marcoceppi:
LGTM :boat:

PR #71, opened 2016-06-14
Non-None default in toggle_state to mean "toggle"

### LINKED ISSUE #70: toggle_state default for active is error-prone
`toggle_state` takes an optional `active` param that lets you specify whether the state should be on or off, instead of it being switched from its previous value.  Having a default value of `None` is problematic because `None` can easily be mistaken for a "truthy" `False` when dealing with relation data, which leads to the state being unexpectedly set.  For example, the following breaks:

```python
conv.toggle_state('{relation_name}.feature_x',
                  active=conv.get_remote('feature_x_available'))
```

This can also be more subtle because `None and <anything>` returns `None` instead of `False` as one might expect.  For example, this also breaks:

```python
name = conv.get_remote('name')
version = conv.get_remote('version')
conv.toggle_state('{relation_name}.ready', name and version)
```

johnsca:
Fixes #70

### DIFF
--- a/charms/reactive/relations.py
+++ b/charms/reactive/relations.py
@@ -30,7 +30,9 @@ from charms.reactive.bus import _load_module
 from charms.reactive.bus import StateList
 
 
-ALL = '__ALL_SERVICES__'
+# arbitrary obj instances to use as defaults instead of None
+ALL = object()
+TOGGLE = object()
 
 
 class scopes(object):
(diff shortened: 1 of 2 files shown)


kwmonroe:
There was some chatter on #juju about not using a string representation for toggle (instead using `object()`).  I'm fine with this as-in, or with that changed.

stub42:
This needs a test, which I suspect is trivial.

johnsca:
Updated with `object()` and test.

stub42:
:boat: :+1: :eggplant:

PR #72, opened 2016-06-15
Fix remote data not being available in -departed

### LINKED ISSUE #53: RelationBase.from_name parameter confusion
RelationBase.from_name has a relation_name parameter, but its doc string says it is a relation id. This is a very important distinction.

### LINKED ISSUE #69: Handling departed hooks in interface layers: relation information is not available.
It is difficult to handle departed relations in interface layers because the relation data is not available.
```conversation.get_remote(key)``` return nothing even though the relation value is available using the relation-get hook tool.

johnsca:
Fixes #69
Also fixes #53


kwmonroe:
LGTM!

PR #73, opened 2016-06-15
Allow @only_once to be used as @only_once()

### LINKED ISSUE #60: only_once help unclear
_From @jacekn on January 27, 2016 17:29_

I was trying to execute certain function only once in my code.

This does not work:
```
@only_once()
def basenode():
    print("in basenode")
```
```
TypeError: only_once() missing 1 required positional argument: 'action'
```

I tried like this but it also did not work:
```
@only_once("basenode")
def basenode():
    print("in basenode")
```
```
AttributeError: 'str' object has no attribute '__code__'
```
Can documentation be clarified to show correct use of this decorator?

_Copied from original issue: juju/charm-tools#94_

johnsca:
Fixes #60

### DIFF
--- a/charms/reactive/bus.py
+++ b/charms/reactive/bus.py
@@ -229,6 +229,7 @@ class Handler(object):
         self._action = action
         self._args = []
         self._predicates = []
+        self._post_callbacks = []
         self._states = set()
 
     def id(self):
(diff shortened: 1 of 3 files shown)


johnsca:
This also resolves the order-dependence issue but it means that `@only_once` can only be used on reactive Handlers.  I don't think anyone is using it outside of handlers, though, anyway.

kwmonroe:
LGTM

PR #74, opened 2016-06-15
Fix logged handler name with @not_unless

### LINKED ISSUE #45: Incorrect handler name logged using not_unless decorator
I'm getting the following in my logs:

2016-01-11 14:58:37 INFO juju-log replication:1: Invoking reactive handler: lib/pypi/charms/reactive/decorators.py:149:_wrapped

It looks like a name isn't being copied from the wrapped function in the not_unless decorator

johnsca:
Fixes #45

### DIFF
--- a/charms/reactive/bus.py
+++ b/charms/reactive/bus.py
@@ -170,12 +170,16 @@ def get_state(state, default=None):
 
 
 def _action_id(action):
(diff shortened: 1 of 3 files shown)


stub42:
Looks good to me. :+1:

PR #75, opened 2016-06-15
Fixed test failure from #72

johnsca:


kwmonroe:
whomever +1'd 73 is a gun jumper.  this fixes 'make test' failures seen with 73.

kwmonroe:
Ok, tests look good *and* the code actually does what we want.

chuckbutler:
@kwmonroe So glad you didn't jump the gun on this one ;)

PR #79, opened 2016-08-04
Allow pluggable alternatives to RelationBase

stub42:
Rather than insisting everything inherits from the fat RelationBase, define a minimal factory type RelationFactory that can be customized and used to build relation instances. The factory defines the two methods required for reactive to acces the relation instances, with the same API as RelationBase. Which makes it backwards compatible since RelationBase and subclasses can serve as their own factories.


johnsca:
Closing in favor of rebased branch in #111

PR #81, opened 2016-08-19
Suppress xtrace for internal functions in charms.reactive.sh

johnsca:
Fixes #80

PR #86, opened 2016-10-21
Add comments to bash magic

johnsca:
As requested in #81


mbruzek:
Thanks for the update @johnsca! this makes more sense to me. +1

PR #100, opened 2017-02-14
The vision of charms.reactive

galgalesh:
This document states the vision of charms.reactive in order to get everyone on the same page.

 - What is the goal of charms.reactive?
 - What is important, where should our focus be?
 - Where do we want to go with the framework?

Let's start the discussion about this!

### DIFF
--- a/VISION.md
+++ b/VISION.md
@@ -3,17 +3,17 @@
 ## Goals
 
  - **Layers** let you build on the work of other charmers.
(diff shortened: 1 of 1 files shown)


galgalesh:
Closing this PR since it is now in the 2.0 branch: https://github.com/juju-solutions/charms.reactive/tree/2.0

PR #106, opened 2017-05-03
Fix cold loading of relation instances

### LINKED ISSUE #105: 0.4.6 breaks charms.openstack (AttributeError: 'NoneType' object has no attribute 'relation_name')
#### Built with charms.reactive 0.4.5: Apr 26 (ok):
https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_amulet_full/openstack/charm-tempest/458466/5/16/index.html

#### Built with charms.reactive 0.4.6: Today (fail):
https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline/openstack/charm-tempest/458466/5/206/index.html

The tempest-0-var-log.tar.bz2  artifact is of interest for both.

johnsca:
From Alex's debugging investigation, it seems that the import loading refactor (#51) made the assumption that handler discovery will always have been run before `RelationBase.from_state()` or `RelationBase.from_name()` are called.  For actions or the CLI, this won't be true, and we may need to attempt to import the appropriate file.

Fixes #105

### DIFF
--- a/charms/reactive/relations.py
+++ b/charms/reactive/relations.py
@@ -14,7 +14,9 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with charm-helpers.  If not, see <http://www.gnu.org/licenses/>.
 
+import os
 import sys
+import importlib
 from inspect import isclass
 
 from charmhelpers.core import hookenv
@@ -25,6 +27,7 @@ from charms.reactive.bus import get_state
 from charms.reactive.bus import set_state
 from charms.reactive.bus import remove_state
 from charms.reactive.bus import StateList
+from charms.reactive.bus import _append_path
 
 
 # arbitrary obj instances to use as defaults instead of None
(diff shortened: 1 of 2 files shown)


charms/reactive/relations.py
```diff
@@ -200,10 +203,14 @@ def _find_impl(cls, role, interface):
...
         # The module has already been discovered and imported.
         module = 'relations.{}.{}'.format(interface, role)
-        if module in sys.modules:
-            return cls._find_subclass(sys.modules[module])
-        else:
-            return None
+        if module not in sys.modules:
+            try:
+                _append_path(hookenv.charm_dir())
+                _append_path(os.path.join(hookenv.charm_dir(), 'hooks'))
```
stub42: I suppose it would be better to have a helper that does these two _append_path calls, shared here and with discover() in bus.py. But that means more lines, and its highly unlikely we will be adding more cruft to sys.path in the future.

tests/test_relations.py
```diff
@@ -98,8 +99,23 @@ def test_from_name(self, _find_impl, join, relation_to_role_and_interface):
         self.assertEqual(res.conversations(), ['conv.join'])
 
     @mock.patch.dict('sys.modules')
+    @mock.patch.object(relations.Conversation, 'join')
+    @mock.patch.object(relations, 'hookenv')
+    def test_cold_import(self, hookenv, conv_join):
+        tests_dir = os.path.dirname(__file__)
+        hookenv.charm_dir.return_value = os.path.join(tests_dir, 'data')
+        sys.modules.pop('relations.hyphen-ated.peer', None)
```
stub42: Thats an odd way of spelling " del sys.modules['relations.hyphen-ated.peer'] ". I thought there was something magic happening here for a moment (is there?)

johnsca: I started with `del`, but that raises an exception if the value isn't in there, while this doesn't (due to the default `None`).

charms/reactive/relations.py
johnsca: Yeah, I had the same thought, that maybe I should factor it into a helper, but then it's only used in two places.  *shrug*

PR #108, opened 2017-06-09
Do not migrate conversations for hooks that are known to have a limited context

tasdomas:
The ```meter-status-changed``` and ```collect-metrics``` hooks have a limited context, meaning they do not have access to many hook environment commands (e.g. relation-ids). This change prohibits the conversation migration part from running on hooks that are known to have a limited context.

### DIFF
--- a/charms/reactive/__init__.py
+++ b/charms/reactive/__init__.py
@@ -17,23 +17,23 @@
 import os
 import sys
 
-from .bus import set_state  # noqa
-from .bus import remove_state  # noqa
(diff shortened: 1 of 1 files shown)


tasdomas:
@johnsca - restricted context hooks and charms.reactive don't mix. There's a lot of layers that depend on things like hookenv.config being available in any hook context. The solution to this will most likely be unrestricting restricted hook contexts. Until then it's best not to use charms.reactive for restricted context hooks.

PR #111, opened 2017-06-26
Allow pluggable alternatives to RelationBase

johnsca:
Rebased version of #79 for merging


johnsca:
For reference: This is enablement for the backwards-compatible transition to [new-style interface layers](https://github.com/juju-solutions/charms.reactive/blob/f8a250e8939c5591866b4da5b0b7d717396a28e2/2.0-proposal/RELATIONS.md) which address the major points of confusion that new authors have with interface layers.

PR #112, opened 2017-06-29
Refactor "state" to "flag"

johnsca:
Rename "states" to "flags" because "flag" better represents the role they perform and how they work.  It also avoids confusion between `status_set` (from `charmhelpers.core.hookenv`) and `set_state` (from `charms.reactive`).

Also refactor the documentation to be better organized and hopefully a bit more clear, particularly to new users.


charms/reactive/relations.py
```diff
@@ -48,19 +49,20 @@ def relation_from_name(relation_name):
         return factory.from_name(relation_name)
 
 
-def relation_from_state(state):
-    """The object used for interacting with relations tied to a state, or None.
+@deprecated.alias('relation_from_state')
+def relation_from_flag(flag):
+    """The object used for interacting with relations tied to a flag, or None.
```
stub42: Alas we need the alias here, because of people being clever with relation_from_state. Its good we are flagging things internal only now.

charms/reactive/relations.py
```diff
@@ -812,22 +814,22 @@ def _migrate_conversations():  # noqa
                     continue
                 value['conversations'].remove(key)
                 value['conversations'].extend(new_keys)
-                set_state(state, value)
+                set_flag(state, value)
 
 
 @cmdline.subcommand()
-def relation_call(method, relation_name=None, state=None, *args):
+def relation_call(method, relation_name=None, flag=None, state=None, *args):
```
stub42: Good you added flag as a new named parameter rather than rename state.

docs/api.rst
```diff
@@ -0,0 +1,17 @@
...
+--------------------------
+
+
+.. toctree::
+
+    charms.reactive.decorators
+    charms.reactive.helpers
+    charms.reactive.relations
+    charms.reactive.bus
+    charms.reactive.cli
```
stub42: Consider only having the actual API in the documentation. Does charms.reactive.bus need to be here encouraging people to use internal only functions? Are these internal only functions? Do we need to separate internals from API better?

docs/index.rst
```diff
@@ -8,20 +8,36 @@ implementations using the reactive pattern.
...
 
-The pattern is "reactive" because you use :func:`@when <charms.reactive.decorators.when>`
-and similar decorators to indicate that blocks of code react to certain conditions,
-such as a relation reaching a specific `state`, a file changing, certain config
-values being set, etc.  More importantly, you can react to not just individual
-conditions, but meaningful combinations of conditions that can span multiple hook
-invocations, in a natural way.
+Juju is an open source tool for modelling a connected set of applications in a
+way that allows for that model to be deployed repeatably and consistently
+across different clouds and substrates.  Juju Charms model the individual
```
stub42: Not sure if model is the correct term here. Maybe they 'implement the model for individual applications' rather than 'model the individual applications'.

docs/index.rst
```diff
@@ -8,20 +8,36 @@ implementations using the reactive pattern.
...
 the appropriate service would be restarted:
 
 .. code-block:: python
 
-    from charms.reactive import when, when_file_changed
+    from charms.reactive import set_flag, clear_flag, when
+    from charms.reactive.helpers import any_file_changed
     from charmhelpers.core import templating, hookenv
 
     @when('db.database.available', 'admin-pass')
```
stub42: We probably want something.admin-pass rather than admin-pass, since we are encouraging a state heirarchy and namespace, and people skimming might get confused by hookenv.config('admin-pass') a few lines down.

docs/structure.rst
```diff
@@ -0,0 +1,43 @@
...
+            └── pgsql
+                ├── interface.yaml
+                ├── peer.py
+                ├── provides.py
+                └── requires.py
+
+The hooks will need to call :func:`reactive.main() <charms.reactive.main>`,
+and the decorated handler blocks can be placed in any file under the ``reactive``
+directory.  Thus, pretty much all of your hooks will end up contain little more
+than:
```
stub42: This is talking as if charm authors create the hooks, whereas the reality is they are pulled in from the base layer. Its nice to explain the mechanism, but it needs to be clearer to charm authors that they are created automatically at charm build time.

docs/structure.rst
```diff
@@ -0,0 +1,43 @@
...
+    from charms.reactive import main
+    main()
+
+The ``relations`` directory will contain any interface layer implementations
+that your charm uses.
+
+If you are `building a charm with layers`_, as is recommended, both the ``hooks``
+and ``relations`` directories will be automatically managed for you by your base
+and interface layers, so you can focus on writing handlers under the ``reactive``
+directory.
```
stub42: Ok, so this is 'remember what we just said? you can ignore it'. Better if it is mentioned earlier, rather than in a coda people might not get to.

tests/test_deprecated.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2014-2016 Canonical Limited.
```
battlemidget: Nitpick, 2014-2017

charms/reactive/bus.py
```diff
@@ -72,38 +77,66 @@ class states(StateList):
...
+    unitdata.kv().update({flag: value}, prefix='reactive.states.')
+    if flag not in old_flags:
+        FlagWatch.change(flag)
+
+
+@cmdline.subcommand()
+@cmdline.no_output
 def set_state(state, value=None):
+    """DEPRECATED Alias of set_flag"""
+    set_flag(state, value)
```
johnsca: Right.  I'm definitely leaning toward just dropping the decorator; I had thought it would save more duplication than it ended up actually doing.

docs/api.rst
johnsca: There are some public API methods in there, like `set_flag`.  It's probably worth breaking them out, as you suggest.

johnsca: Though, we also import them into the top-level `charms.reactive` module; maybe we should just commit to having that top-level module represent the public API and document from there.

docs/index.rst
```diff
@@ -8,20 +8,36 @@ implementations using the reactive pattern.
...
-conditions, but meaningful combinations of conditions that can span multiple hook
-invocations, in a natural way.
+Juju is an open source tool for modelling a connected set of applications in a
+way that allows for that model to be deployed repeatably and consistently
+across different clouds and substrates.  Juju Charms model the individual
+applications, their configuration, and the relations between them and other
+applications.
+
+In order for the charm to know what actions to take, Juju informs it of
+life-cycle events in the form of hooks.  These hooks inform the charm of things
```
AdamIsrael: We don't need to spaces after a period. Perhaps we can build a style guide to help document best/recommended practices. It'd be nice if we could also do a "lint" against the docs and a style guide.

docs/interface-layers.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Reactive charms work best as a part of layered charms, and the reactive pattern
+enables the use of interface layers which encapsulate the communication protocol
+over a Juju interface when two applications are related together.  By
+encapsulating this protocol, it ensures that two applications that support the
+same interface will always be able to speak to each other correctly.  It also
+provides a well-defined and documented API for the charm author to use, without
+having to get mired in the details of relation keys, data encoding, or what
+order the communication has to proceed in.
+
```
AdamIsrael: It might be nice to add an ascii diagram here to visualize how the layers work together.

docs/structure.rst
AdamIsrael: It'd be nice to document more of what a reactive charm might contain, such as metrics.yaml, or actions.yaml and actions/

docs/structure.rst
AdamIsrael: I think it's a safe guess that if the user is reading these docs, they are or planning to write a charm using layers.

tests/test_deprecated.py
AdamIsrael: Would it be more useful to have singular LICENSE and COPYRIGHT files in the root of the repository, and replace the copyright statement in the header of every file with a reference to that? Otherwise, we end up having to touch every file, every year, to keep it up to date, rather than just the two in the root. We'd need to research the license requirements and see if there are any recommended best practices for this.

docs/index.rst
johnsca: Considering these docs are intended to be read in their rendered, whitespace-collapsed form, does this really matter?

johnsca:
Updated per reviews.  New docs can be viewed online at: https://charmsreactive.readthedocs.io/en/refactor-state-flags/

docs/structure.rst
```diff
@@ -1,43 +1,139 @@
...
+  └──────┬──────┘ └──────┬────┘ └─────────┬──────┘
+         └───────────────┼────────────────┘
+                  ┌──────┴────────┐ ┌─────────────────┐
+                  │ layer:apache2 │ │ interface:mysql │
+                  └──────┬────────┘ └────────┬────────┘
+                         └─────────┬─────────┘
+                               ┌───┴───┐
+                               │ myapp │
+                               └───────┘
+
```
battlemidget: Thanks, this is very helpful

docs/structure.rst
```diff
@@ -1,43 +1,139 @@
...
+layer to provide common behavior, using documented flags and method calls to
+communicate with those layers.
+
+A charm layer consists, at a bare minimum, of the following files:
+
+* ``metadata.yaml``: This file contains information about the charm, such as the
+  charm name, summary, description, maintainer, and what relations the charm
+  supports.
+* ``layer.yaml``: This file indicates what other layers this charm builds upon.
+* ``reactive/<charm-name>.py``: This file, where ``<charm-name>`` is replaced by the
```
battlemidget: Maybe just set this as `<charm_name>.py` for additional clarity

docs/structure.rst
```diff
@@ -1,43 +1,139 @@
...
-and ``relations`` directories will be automatically managed for you by your base
-and interface layers, so you can focus on writing handlers under the ``reactive``
-directory.
-
-.. _`Building a Charm with Layers`: https://jujucharms.com/docs/stable/authors-charm-building
+    ├── icon.svg
+    ├── config.yaml
+    ├── layer.yaml
+    ├── reactive/
+    │   └── myapp.py
```
battlemidget: Maybe show this as `my_app.py` to keep the examples the same

docs/structure.rst
```diff
@@ -1,43 +1,139 @@
...
+.. _layers: https://jujucharms.com/docs/stable/authors-charm-building
+__ layers_
+.. _`Charm Store`: https://jujucharms.com/
+
+
+Base Layers
+-----------
+
+Base layers provide functionality that is common across several charms. These
+layers should provide a set of handlers in ``reactive/<layer-name>.py`` which
```
battlemidget: Same with just show it as underscores `<layer_name>.py`

docs/structure.rst
```diff
@@ -1,43 +1,139 @@
...
+will set additional flags that will drive behavior in the charm layer. They may
+also include a Python module in ``lib/charms/layer/<layer-name>.py`` which can
```
battlemidget: Underscores here as well

docs/structure.rst
```diff
@@ -1,43 +1,139 @@
-
 Structure of a Reactive Charm
------------------------------
+=============================
+
+A reactive charm should be built using layers, with the "top" layer being called
```
ajkavanagh: Typo?  replace "should" with "must" or "should be" with "is".  It's not optional (really) as layer-basic is part of charms.reactive (in essence) ... I guess, unless the charm author wants to completely replace the layer-basic behaviour?

docs/index.rst
AdamIsrael: I would argue yes, if only because it's an assumption that the docs will only be read digitally rendered. One space after a period is [best practice](http://www.quickanddirtytips.com/education/grammar/two-spaces-after-a-period), like spaces instead of tabs and vim instead of emacs. Okay, kidding about that last one. Kind of.

I don't think the spaces are a blocker. I think there's some doc cleanup and standardization that needs doing in general. That's currently tracked in #114.

PR #119, opened 2017-07-14
Only execute matching hooks in restricted context.

tasdomas:
Some hooks (collect-metrics, meter-status-changed) are executed in a restricted context. Most hook tools are not available in such contexts, so generic flag handlers should not be executed by charms.reactive. However, we still want layered charms to be able to react to these hooks. A separate main method is added for this purpose.

This method (```main_restricted```) will only execute matching hooks (not state handlers) in a restricted context.

### DIFF
--- a/charms/reactive/__init__.py
+++ b/charms/reactive/__init__.py
@@ -28,6 +28,7 @@ from .flags import remove_state  # noqa  DEPRECATED
 from .flags import toggle_state  # noqa  DEPRECATED
 from .flags import is_state  # noqa  DEPRECATED
 from .flags import all_states  # noqa  DEPRECATED
+from .flags import get_states # noqa  DEPRECATED
 from .flags import any_states  # noqa  DEPRECATED
 from .relations import scopes  # noqa
 from .relations import RelationBase  # noqa
@@ -41,6 +42,8 @@ from .decorators import when_not_all  # noqa
 from .decorators import not_unless  # noqa
 from .decorators import only_once  # noqa
 from .decorators import when_file_changed  # noqa
+from .decorators import collect_metrics # noqa
+from .decorators import meter_status_changed # noqa
 
 from . import bus
 from . import relations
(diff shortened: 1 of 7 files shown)


johnsca:
@tasdomas Sorry for the delayed response; sprints and whatnot.

I had already been talking with some of the other devs about these types of hooks, particularly in the context of `update-status` (see juju-solutions/layer-basic#100), and I think this PR is in the right general direction.  However, there are a couple of issues I have with the specifics.

First, I think we should move away from the base layer having implementation for things that represent fundamental aspects of how the Juju model works for charms, such as the fact that certain hooks have different types of context or are expected to perform different types of actions from "normal" hooks.  Thus, I would rather see the existing reactive main have this behavior folded in (as well as pulling in the other behavior currently in layer-basic, such as the aforementioned PR and the `config.X` flag management), rather than having a separate main with the base layer having to figure out which to call when.  That is to say, I would like to get back to the base layer being responsible solely for getting the reactive framework and charm code bootstrapped on the system in question.  (This will become more relevant in the future when we potentially have a different base layer for, e.g., Windows or Centos.)

Second, I **really** want to strongly deprecate `@hook`.  The usage of that decorator has led to significant issues with the ability to upgrade from non-reactive to reactive charms, and it also carries forward the potential for hook invocation order dependency errors in charms which was the main reason we created reactive style charms in the first place.  So, I'd like to extend the model that we use with the `config.changed` flag(s) to also handle cases like this (as well as `update-status` and the relation events).

tasdomas:
@johnsca - thanks for the review. I've updated the PR. 

I'll wrap the hook files into a separate layer and I'm closing https://github.com/juju-solutions/layer-basic/pull/101/files

tasdomas:
@johnsca, @stub42  - PTAL

charms/reactive/bus.py
```diff
@@ -342,6 +344,13 @@ def _invoke(to_invoke):
                     break
         FlagWatch.commit()
 
+    # When in restricted context, only run hooks for that context.
+    if restricted:
+        unitdata.kv().set('reactive.dispatch.phase', 'restricted')
+        hook_handlers = _test(Handler.get_handlers())
+        _invoke(hook_handlers)
+        return
```
stub42: An alternative to creating a new phase would be to just reuse the existing hook phase. A new phase is likely cleaner, but might derail some of the inflight reactive 2.0 work.

charms/reactive/decorators.py
```diff
@@ -222,3 +223,25 @@ def only_once(action=None):
     handler.add_predicate(lambda: not was_invoked(action_id))
     handler.add_post_callback(partial(mark_invoked, action_id))
     return action
+
+
+def collect_metrics():
+    """
+    Register the decorated function to run for the collect_metrics hook.
+    """
```
stub42: If you reused the existing hook phase, these two decorators would not be needed. You could just use `@hook('collect_metrics')`

charms/reactive/__init__.py
```diff
@@ -59,29 +62,36 @@ def main(relation_name=None):
...
     try:
         bus.discover()
-        hookenv._run_atstart()
-        bus.dispatch()
+        if not restricted_mode:  # limit what gets run in restricted mode
+            hookenv._run_atstart()
+        bus.dispatch(restricted=restricted_mode)
     except SystemExit as x:
         if x.code is None or x.code == 0:
             hookenv._run_atexit()
```
tasdomas: Good point. I'll fix this.

charms/reactive/decorators.py
tasdomas: I was told that `@hook` decorators have fallen out of favor.

tasdomas:
@stub42 @johnsca  - do you have any suggestions as to the direction this should move in? We have partners waiting for these fixes and it looks like this has been marinating long enough. Thanks!

stub42:
I'm fine with this.

If use of decorators to mark the new handlers is a sticking point, we can document them as experimental. This gives us the leeway to change things if needed with the ongoing charms.reactive updates and migration from states to flags.

tasdomas:
@stub42 - thanks!
@johnsca - what do you think?

P.s. CI currently fails due to a cyclomatic complexity check - do you think the threshold could be adjusted?

johnsca:
@tasdomas My inclination is to avoid new decorators whenever possible, and these have all of the same issues as `@hook` in that they're tied to a specific hook execution and don't really fit with the motivation behind reactive.  However, I also think that, at least, the `collect-metrics` hook is actually more like actions in that they express specific intent as to what code path should be invoked and that it should return a result.  That said, I think we're still missing the feature I would consider necessary to do this correctly (stable handlers), so I'm ok with moving forward with this and refactoring as needed in the future.  This really should only affect one (albeit base) layer and should be easy enough to make backwards compatible when / if we do refactor it.

PR #101, opened 2017-02-14
charms.reactive 2.0

galgalesh:
This PR is to discuss the 2.0 changes. Initially just design/docs, later actual code.

To see the actual README rendered by github: https://github.com/juju-solutions/charms.reactive/tree/2.0/2.0-proposal

*I like the "discussion flow" of a PR, but everyone should be able to change the proposal, hence this setup. Push to the 2.0 branch to edit this PR.*

PR #121, opened 2017-08-16
Add flag triggers

johnsca:
As discussed in #101.

### DIFF
--- a/charms/reactive/__init__.py
+++ b/charms/reactive/__init__.py
@@ -20,6 +20,7 @@ import sys
 from .flags import set_flag  # noqa
 from .flags import clear_flag  # noqa
 from .flags import toggle_flag  # noqa
+from .flags import register_trigger  # noqa
 from .flags import is_flag_set  # noqa
 from .flags import all_flags_set  # noqa
 from .flags import any_flags_set  # noqa
--- a/charms/reactive/flags.py
+++ b/charms/reactive/flags.py
@@ -50,6 +50,11 @@ def set_flag(flag, value=None):
     unitdata.kv().update({flag: value}, prefix='reactive.states.')
     if flag not in old_flags:
         FlagWatch.change(flag)
+        trigger = _get_trigger(flag)
+        for flag_name in trigger['set_flag']:
+            set_flag(flag_name)
+        for flag_name in trigger['clear_flag']:
+            clear_flag(flag_name)
 
 
 @cmdline.subcommand()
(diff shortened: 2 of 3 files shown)


johnsca:
[(2) triggers spike](https://trello.com/c/W9oXSe5q/2-2-triggers-spike)

ajkavanagh:
I (almost) hate to say it, but I actually think this will be a very useful addition to charm writing. +1

johnsca:
Updated the description with a recap of the motivation.

@ajkavanagh I know the discussion sort of trailed off with you perhaps still having concerns as to whether it was necessary, but I do think there is the one case above that can't really be solved without something new being added to reactive.  This seems like the smallest change that solves the use-case and isn't likely to introduce other side-effects.

I'm happy to continue to discuss the pros and cons here, though.

johnsca:
When we have stable handlers (i.e., handlers are not reinvoked during a hook if a given flag was previously set but has not changed), then triggers should also be stable (they shouldn't be retriggered unless the `when` flag is new or explicitly updated).  And if they are triggered, the `set_flag` variant should force the dependent flag to be updated even if already set.

Triggers should not affect any handlers already in the execution queue.  (I also want to remove the queue interruption that currently happens with `clear_flag`.)

I'm not sure if there's any need for a `when_not` (`when_cleared`?) variant of triggers, but I suspect not.

simonklb:
I do agree that this is a quick and simple fix for the problem described above. However, I'm still firmly convinced that introducing dependencies and execution-order in the flag-model (which also would solve this problem) is the way to go if the goal is to make charm-development easier to adopt.

Even though that would be somewhat more extensive it wouldn't necessarily force you to deprecate the current state of affairs and the benefits of it would be well worth it. It would make charm authors declare everything about a flag's behavior in a single place and I think it would be much more straightforward to visualize the execution of handlers.

That being said, I don't want to derail your current milestones. So if you want to go ahead with this I'm not against it. I just don't want to stay silent on the matter as this is something that I think is really important for the reactive framework.

ajkavanagh:
@johnsca I meant that I really do approve of this.  I've had a few use cases where this will come in very, very useful.

I'm also thinking that a 'transient' flag would be useful; i.e. it exists only for the duration of the current hook invocation and then disappears.  Combined with this change, you could do some nice things:

1. Set the transient in the interface.
2. trigger off that flag (to get trigger-flag) in the charm layer to get a 'persistent until I clear it' flag for a handler.
3. when the handler has completed successfully (i.e. made it to the end), the handler clears the trigger-flag.

This would mean that we have fault tolerant 'changed' flags that survive a 'crash', but then 'recover' when juju retries the hook. (I know that this isn't good practice; but it's about trying to be more fault tolerant).

Pseudo code:

```python
# in interface
@hook('some-name-relation-{joined,changed}')
def joined(self):
    ...
    if changed:
         self.set_transient_flag('some-flag')

# in charm layer

flags.register_trigger(when='some-flag', set_flag='myhandler-some-flag')

@when('myhandler-some-flag')
@when('some-relation.joined'')
def do_myhandler():
    ... do something that might crash (as we can never be sure ...)
    flags.remove_flag('myhandler-some-flag')
```

P.S. I've done the 'transient' flag by using a `hookenv.atexit(...)` to drop the flag at the end of hook execution; it only works because `charms.reactive` also sets up an `atexit(...)` handler that flushes the `kv()` store and it runs _last_ because it gets registered before any other atexit(...) hooks :smile:

PR #122, opened 2017-08-31
Added docs for triggers

### LINKED ISSUE #97: Link states together
This might be a more permanent solution to the issues @stub42 is facing with the apt layer. A simplification of the issue in question:

```python
@when("config.changed.packages")
@when("apt.sources.configured")
def  install_packages():
    # ...

@when("config.changed.sources")
def configure_sources():
    # ...
```

The issue here is that `configure_sources` has to be run before `install_packages` if both `config.changed.packages` and `config.changed.sources` are set.

I'd like to find a solution that isn't tied to implementation details of the reactive framework (ie. the fact that a "run" starts as part of a hook). Nor to the technical details of this issue. An [`@preflight` decorator](https://github.com/juju-solutions/charms.reactive/pull/20) solves this issue because config cannot change during an reactive run. However, other stuff can change such as files changing or states being added and removed. We also want to cover those cases.

I don't think the reactive framework can currently solve this issue in a clean way.

A solution for this problem that fits the reactive approach better is to be able to link states together so that state changes can trigger other state changes without the need for a handler to run.

```python
# Every time `config.changed.sources` gets set, then `apt.sources.configured` gets removed
# immediately; before running the next handler.
reactive.if_this_then_remove('config.changed.sources', 'apt.sources.configured')

@when("config.changed.packages")
@when("apt.sources.configured")
def  install_packages():
    # ...

@when("config.changed.sources")
def configure_sources():
    # ...
    set_state('apt.sources.configured')
```

I think that solves @stub42's issue. Thoughts?

johnsca:
This finishes out the triggers feature and resolves #97.

### DIFF
--- a/docs/triggers.rst
+++ b/docs/triggers.rst
@@ -1,0 +1,54 @@
+
+Coupling Flags with Triggers
+----------------------------
+
+In general, it is best to be explicit about setting or clearing a flag.  This
+makes the code more maintainable and easier to follow and reason about.
+However, rarely, due to the fact that handlers for a given flag are
+independent and thus there are no guarantees about the order in which they may
+execute, it is sometimes necessary to enforce that two flags must be set at
+the same time or that one must be cleared if the other is set.
+
+As an example of when this might be necessary, consider a charm which provides
+two config values, one that determines the location from which resources should
+be fetched, with a default location provided by the charm, and another which
+indicates that a particular feature be installed and enabled.  If the charm is
+deployed and fetches all of the resources, it might set a flag that indicates
+that all resources are available and any installation can proceed.  However, if
+both resource location and feature flag config options are changed at the same
+time, the handlers might be invoked in an order that causes the feature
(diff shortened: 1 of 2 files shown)


johnsca:
Test is failing due to [lp:1714318](https://bugs.launchpad.net/juju/+bug/1714318) and is unrelated to the PR (which is docs only)

johnsca:
[(3) Docs for triggers](https://trello.com/c/RvFr0U6Y/46-3-docs-for-triggers)

docs/triggers.rst
```diff
@@ -0,0 +1,54 @@
+
+Coupling Flags with Triggers
+----------------------------
+
+In general, it is best to be explicit about setting or clearing a flag.  This
+makes the code more maintainable and easier to follow and reason about.
+However, rarely, due to the fact that handelers for a given flag are
```
stub42: handlers, not handelers

PR #123, opened 2017-09-14
Start of Endpoint implementation for easier interface layers

### LINKED ISSUE #61: RelationBase.conversation() returns an arbitrary SERVICE or UNIT scoped conversation
A service can be related to another service multiple times. To cope with this, conversations are namespaced using the relation id (the only key), and we end up with one SERVICE scoped conversation per relation

The RelationBase.conversation() classmethod only has a single scope parameter, and can only return one of the several SERVICE scoped conversations. The rest are inaccessible.

The same issue also occurs for UNIT scoped conversations.

### LINKED ISSUE #88: 'conversation.set_state' is too ambiguous, should use 'conversation.add_to_state'
After debugging for 2.5 hours together with @galgalesh, I finally found that our code was completely wrong due to our understanding of the following function.

    conversation.set_state('{relation_name}.state')

Our understanding was that this adds that state to the conversation, while in fact it is actually the other way around; This function adds the conversation to that state.

Suddenly everything makes sense:

- Relationship states contain conversations because you create them by adding a conversation to a state.
- Normal states don't contain relationships because you create them by creating the state.
- Removing a relationship from a state 
- A relationship state without conversations gets removed.
- An "x.broken" state contains all the conversations that were "broken".
- etc...

To prevent other people from tripping over this, it might be best to alias `conversation.set_state` to `conversation.add_to_state('{relation_name}.state')`. That would make it more clear what this function does exactly. Similarly, alias `conversation.remove_state` to `conversation.remove_from_state`. You are not adding or removing a state to/from a conversation, you are adding and removing a conversation to/from a state.

### LINKED ISSUE #102: Layered interface layers
We need a way to share code between interface layers.

# `rest-api` extends `http`

Consider a database WololoDB® with an HTTP REST API. The charm for WololoDB should implement the generic `http` interface to connect to reverseproxies, loadbalancers, https endpoints etc. The charm should also implement a specific `wololodb` interface for clients who want to talk to the database. So we have two interface layers:

- `http` sends the hostname and the port and tells Juju that it can connect to other charms implementing the `http` interface.
- `wololodb` sends the hostname and port and tells Juju that it can connect to other charms implementing the `wololodb` interface.

Although both interfaces are identical, they need to be separate so that Juju knows that a `wololodb` client can only connect to a WololoDB server, not all HTTP servers.

This maps very cleanly to layers, the `wololodb` interface uses the `http` interface as base interface layer.

Whatever we come up with for reactive 2.0 should enable this kind of layered approach to interfaces.

# Totally epic

Juju should understand the interface inheritance. The HAProxy charm is a great use-case for this. The `reverseproxy` relationship has two "modes", the default mode compatible with the normal `http` interface, and an advanced mode where more info is shared over the relationship.

This would result in two interfaces

 - `http`
 - `reverseproxy` which `includes: ['http']`

When an operator connects a charm to haproxy, Juju knows that the `reverseproxy` relationship is preferred over the `http` relationship if both charms implement that interface.

### LINKED ISSUE #107: scopes use SERVICE vs APPLICATION
I'm trying to get the docs updated and noticed that the code example can't update the language as the code is using the SERVICE keyword vs APPLICATION.

johnsca:
Beginning of the implementation for the [proposal][] for new style interface layers using the `Endpoint` base, eschewing the reliance on `@hook` in favor of `@when` everywhere, and removal of the "conversations" metaphor.  This implements up to and including the "Managed Flags" section.

fixes #107
fixes #88
fixes #61
fixes #102 

[proposal]: https://github.com/juju-solutions/charms.reactive/blob/2.0a1/2.0-proposal/RELATIONS.md


charms/reactive/altrelations.py
```diff
@@ -41,9 +44,68 @@ def from_name(cls, relation_name):
...
+
+    @classmethod
+    def from_flag(cls, flag):
+        parts = flag.split('.')
+        if len(parts) < 3 or parts[0] != 'relations':
+            return None
+        return cls.from_name(parts[1])
+
+
+hookenv.atstart(Endpoint._startup)
```
stub42: Endpoint._startup should not be called in 'restricted' mode, or metrics and similar hooks are going to become heavyweight.

charms/reactive/relations.py
```diff
@@ -267,6 +272,16 @@ def changed(self):
     recommended that this be used only with :attr:`scopes.GLOBAL` scope.
     """
 
+    @classmethod
+    def _startup(cls):
+        # update data to be backwards compatible after fix for issue 28
+        _migrate_conversations()
```
stub42: Similarly here, if _migrate_conversations isn't usually a noop we shouldn't call it in restricted_mode.

charms/reactive/altrelations.py
johnsca: `_run_atstart` isn't called in restricted mode, so it won't be.

johnsca:
[(3) RelationBase: Relation and Unit collections](https://trello.com/c/KTjmlUnE/64-3-relationbase-relation-and-unit-collections)

johnsca:
[(3) RelationBase: Sending and Receiving Relation Data](https://trello.com/c/Eacc62th/65-3-relationbase-sending-and-receiving-relation-data)

charms/reactive/bus.py
```diff
@@ -65,18 +75,19 @@ class Handler(object):
     _CONSUMED_FLAGS = set()
 
     @classmethod
-    def get(cls, action):
+    def get(cls, action, suffix=None):
         """
         Get or register a handler for the given action.
 
         :param func action: Callback that is called when invoking the Handler
         :param func args_source: Optional callback that generates args for the action
```
battlemidget: Does this need to be updated to reflect the name of the function argument?

charms/reactive/bus.py
```diff
@@ -93,14 +104,14 @@ def clear(cls):
         """
         cls._HANDLERS = {}
 
-    def __init__(self, action):
+    def __init__(self, action, suffix=None):
         """
         Create a new Handler.
 
         :param func action: Callback that is called when invoking the Handler
         :param func args_source: Optional callback that generates args for the action
```
battlemidget: Here to?

charms/reactive/helpers.py
```diff
@@ -17,6 +17,7 @@
 import re
 import json
 import hashlib
+from types import SimpleNamespace
```
battlemidget: best module ever

kwmonroe:
+1 on the recent docstring commit -- really nice job explaining wth is going on @johnsca!

johnsca:
[(2) Endpoint: Documenting Interface Layers](https://trello.com/c/v9A3M9uV/66-2-endpoint-documenting-interface-layers)

jamesbeedy:
@johnsca this is really great. One question, how might a peer relation take advantage of this new approach?

galgalesh:
I finally found time to look at this PR, I'm really excited about it!

I pushed some docstring changes, please let me know if I'm writing stuff that's incorrect. I feel that "publishing" is a good description of the behavior of `relation-set`, so I used that to describe the behavior of the `send` dictionary. What are your thoughts about changing the name of `send` to `to_publish`? I feel that explains its behavior better and it addresses the concern we had that `send` is not true to what actually happens behind the scenes.

docs/charms.reactive.decorators.rst
```diff
@@ -3,6 +3,38 @@ charms.reactive.decorators
...
+The only exceptions to this are enpoint handlers, handlers that are instance
+methods of :class:`~charms.reactive.altrelations.Endpoint`: they get the
+endpoint object as the `self` argment.
+
+If a handler needs to use an older interface using the legacy
+:class:`~charms.reactive.relations.RelationBase`, or if you're uncertain which
+is used by an interface layer, you can also obtain an instance using
+:func:`~charms.reactive.relations.relation_from_flag`.
+
+For backwards compatibility, some decorators can pass instances of these
```
galgalesh: How do we handle the case when a handler doesn't accept instances but the decorator wants to pass one? (a "new" handler reacting to a flag of an "old" interface)

docs/charms.reactive.decorators.rst
```diff
@@ -3,6 +3,38 @@ charms.reactive.decorators
...
+Regular handlers should not accept any arguments. When a handler needs to use a
+(relationship) :class:`~charms.reactive.altrelations.Endpoint`, it can access
+the endpoint object via the :data:`~charms.reactive.helpers.context` namespace.
+The only exceptions to this are enpoint handlers, handlers that are instance
+methods of :class:`~charms.reactive.altrelations.Endpoint`: they get the
+endpoint object as the `self` argment.
+
+If a handler needs to use an older interface using the legacy
+:class:`~charms.reactive.relations.RelationBase`, or if you're uncertain which
+is used by an interface layer, you can also obtain an instance using
```
galgalesh: What do you mean with

>  if you're uncertain which is used by an interface layer

?

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+        """
+        if self._units is None:
+            self._units = CombinedUnitsView([
+                RelatedUnit(self, unit_name) for unit_name in
+                sorted(hookenv.related_units(self.relation_id))
+            ])
+        return self._units
+
+    @property
+    def send_json(self):
```
galgalesh: What do you think about refactoring `send_json -> send` and `send -> send_raw`? This will make it more clear that the json variant should be used unless you need lower-level access to be compatible with an existing interface.

(`to_publish` and `to_publish_raw` if you agree on that change too)

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+                hookenv.atexit(relation._flush_data)
+
+    def __init__(self, relation_name, relation_ids=None):
+        self._relation_name = relation_name
+        self._relations = KeyList(map(Relation, relation_ids or []),
+                                  key='relation_id')
+        self._all_units = None
+
+    @property
+    def relation_name(self):
```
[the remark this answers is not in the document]
johnsca: My thinking here (and for the `Endpoint` vs `EndpointInterface` above) was that the class represented the endpoint interface but specific instances represent actual relations (since they have an associated ID).  But I'm also feeling like that distinction might be too esoteric, and really the name is associated with the endpoint itself, not a specific relation.  So +1 on this and the above class name change.

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+        return self._relations
+
+    @property
+    def joined(self):
+        """
+        Whether this endpoint has remote applications attached to it.
+        """
+        return len(self.relations) > 0
+
+    def flag(self, flag):
```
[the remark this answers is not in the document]
johnsca: This was really more of a helper for populating the `{endpoint_name}` (I'm inclined to change that to match the method name, discussed above) portion in a consistent way, and to provide a shortcut if you didn't want to type that bit out all the time (probably a bad idea?).  I'm not sure how it would be used to document public flags for the layer since they happen at runtime; can you explain your thoughts there more?

I had also toyed around with the idea of having some class-level structure to pre-define the flags, but you still couldn't populate the `{endpoint_name}` portion until runtime, so I'm not sure how we could make that work.  But that would certainly give us a great way to be able to add doc strings for the flags themselves, perhaps even integrated into the definition itself.

charms/reactive/altrelations.py
johnsca: +1 on both.

docs/charms.reactive.decorators.rst
johnsca: Well, old-style relation instances aren't populated into the global context (because they depend on which specific flag you're handling), so you have to use `relation_factory` (or the terrible ambiguous handler args) to get access to them.  And whether an interface layer uses `EndpointInterface` or `RelationBase` is an implementation detail that the charm author shouldn't care about.  This will be problematic during the transition, so I'm not sure how to handle it gracefully.

docs/charms.reactive.decorators.rst
johnsca: I added introspection logic to the invocation code so that, if the handler doesn't take any arguments, it will do the right thing and not pass them.  You'd then have to use `relation_factory` or `context.endpoints` to access them, depending on how they're implemented (ugh).  Though, the decorators will also pass the args through for `EndpointInterface` flags as well, so we'll probably end up having to continue to do that until everything is converted and then we'll still have the huge amount of inertia to move away (ugh).

johnsca:
@jamesbeedy I answered your question on IRC, but for posterity here...

Peer relations are handled exactly the same as other relations, with the only difference being you only need one class, in a `peers.py` file, rather than two (in `requires.py` and `provides.py`).  The class itself will be the same, just publishing data and setting flags based on what data has been published.  Also, peer relations are guaranteed to only ever have a single relation ID, which makes the helper methods (like `all_units`) particularly useful for them.

jamesbeedy:
This is great. Thanks @johnsca

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+
+        # and of course you an access them by individual unit
+        assert endpoint.relations['endpoint:1'].units['unit/1'].received['key0'] \
+                == 'value0_1_1'
+
+    """
+    def __init__(self, items):
+        super().__init__(sorted(items, key=lambda i: (i.relation.relation_id,
+                                                      i.unit_name)),
+                         key='unit_name')
```
stub42: This is sorting by relation id as a string, which will put db:10 before db:2. Similarly the unit_name, which you claim later is order by unit number. Consider key=lambda i: (int(':'.split(i.relation.relation_id)[1]), int('/'.split(i.unit_name)[1]))

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+        return self.received_json.data
+
+    @property
+    def received_json(self):
+        """
+        Combined :class:`~charms.reactive.altrelations.JSONUnitDataView` of the
+        data of all units in this list, with automatic JSON decoding.
+        """
+        if not hasattr(self, '_data'):
+            # NB: units are reversed so that lowest numbered unit takes precedence
```
stub42: yer, if you want numerical order you need to do it above.

charms/reactive/bus.py
```diff
@@ -65,18 +75,19 @@ class Handler(object):
     _CONSUMED_FLAGS = set()
```
stub42: Not sure why you want _CONSUMED_FLAGS here instead of _consumed_flags. All-caps is generally reserved for constants in Python. Similarly, _HANDLERS below.

charms/reactive/bus.py
```diff
@@ -40,21 +40,31 @@ def __init__(self, path):
...
+        action.__code__.co_firstlineno,
+        action.__code__.co_name,
+    ]
+    if suffix is not None:
+        parts.append(suffix)
+    return ':'.join(map(str, parts))
 
 
-def _short_action_id(action):
+def _short_action_id(action, suffix=None):
```
stub42: Should suffix be optional? Its an internal method, so making it a required argument will help catch any call sites we missed.

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+    The reactive framework automatically creates endpoint objects in the
+    :data:`~charms.reactive.helpers.context` namespace. Handlers can access the
+    endpoint objects from there, even if that endpoint doesn't have a joined
+    relation. The name of the endpoint object is based on the endpoints
+    configured in `metadata.yaml`. Note that endpoints will only be created
+    **if the interface layer of that endpoint is included in the `layer.yaml`
+    file.**
+
+    Endpoint handlers can iterate over the list of joined relations for an
+    endpoint via the :meth:`~charms.reactive.altrelations.Endpoint.relations`
```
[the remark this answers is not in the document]
galgalesh: This is correct markup and it gets formatted as `relations`, the full path isn't shown in the text but is used to create a url to those docs.. PS: HACKING.md explains how you can build the docs locally.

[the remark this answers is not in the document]
galgalesh: But `:attr:` is better because then it doesn't show the `()` after the name. Fixed.

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+        """
+        return self._relation_name
+
+    @property
+    def relations(self):
+        """
+        Collection of `Relation`s that are established for this `Endpoint`.
+
+        This is a `KeyList`, so it can be iterated and indexed as a list,
+        or you can look up relations by their ID.  For example::
```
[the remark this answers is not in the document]
johnsca: I do think we should push back on having relation IDs reported as plain numbers, and I'm +1 to normalizing it in the framework if Juju does report them that way.

On the other hand, though, from the charm author's perspective, I would prefer relation IDs to be an implementation detail that they don't need to worry about, as much as possible.

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+    """
+
+    @classmethod
+    def from_name(cls, relation_name):
+        return getattr(context.endpoints, relation_name, None)
+
+    @classmethod
+    def from_flag(cls, flag):
+        if '.' not in flag:
+            return None
```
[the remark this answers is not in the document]
johnsca: This is just to match how `RelationBase.from_flag` and `MinimalRelationBase.from_flag` return `None` when a relation instance can't be derived from a flag.  The only difference being that `EndpointInterface.from_flag` could return an instance even if the flag isn't set.  +1 to adding a comment.

charms/reactive/altrelations.py
```diff
@@ -41,9 +49,574 @@ def from_name(cls, relation_name):
...
+    *Note that these flags can be used in the decorators of all handlers, not
+    just endpoint handlers, although this should be done with caution.*
+
+    The reactive framework automatically creates endpoint objects in the
+    :data:`~charms.reactive.helpers.context` namespace. Handlers can access the
+    endpoint objects from there, even if that endpoint doesn't have a joined
+    relation. The name of the endpoint object is based on the endpoints
+    configured in `metadata.yaml`. Note that endpoints will only be created
+    **if the interface layer of that endpoint is included in the `layer.yaml`
+    file.**
```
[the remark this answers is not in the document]
johnsca: I'm starting to wonder about the viability of the `context` pattern for this.  As you point out, the name is a bit generic, it requires magic translation of the endpoint name to a valid attribute name, and it ends up exposing the internal implementation (whether the interface layer uses `EndpointInterface` or `RelationBase`) by virtue of whether a given endpoint can be accessed via the `context` or not (although this last is only an issue during the transition, that transition is likely to be long).

We'll probably need to go back to recommending `relation_from_flag` only.

charms/reactive/altrelations.py
johnsca: Good catch, thanks.

johnsca:
Per @galgalesh's request, since the http interface is used as the example in the docs, I updated the http interface to use Endpoint in this PR: juju-solutions/interface-http#10

Issue #118, opened 2017-06-30
How to debug a reactive charm


AdamIsrael:
Stolen from [Ask Ubuntu](https://askubuntu.com/questions/804465/juju-charms-how-to-run-and-debug-reactive-charm):

> I'm developing a juju charm in python using a reactive pattern. All my hooks are called using decorators provided by 'basic' base layer. When trying to deploy the charm I encountered an error. In order to solve the error I used 'juju debug-hooks' and 'juju dhx -r' (dhx debugging plugin) but I can't find a way to run charm code using python following reactive pattern... As I understand, (when using dhx for debugging for example) the charm execution is returned to the state exactly before the hook that caused an error and the developer is logged in into a remote debugging session. It is on the developer to run hooks. Normally these hooks would be under /hooks directory and would be run by just executing them. This isn't possible using reactive pattern, since there is no entry point for hooks but everything is just a cascade of reactions to states...

I think this question raises two good points:
1. We should maintain a section for frequently asked questions that can occasionally be updated based on feedback on Ask Ubuntu, #juju, or the mailing lists.
2. In a case of this question, we should link to the relevant Juju documentation about debugging a charm in general, and then talk about the specific case of debugging a reactive charm.

PR #129, opened 2017-10-05
Added FAQ section to docs

galgalesh:
- Added how to run a reactive charm in debug. Fixes https://github.com/juju-solutions/charms.reactive/issues/118
- Added `HACKING.md` with explanation of how to build docs and browse locally

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,30 @@
+# Hacking on Docs
+
+Install dependencies:
+
+```bash
+sudo apt install python3-sphinx
+```
+
+To build the html documentation:
+
+```bash
+make docs
+```
+
+To browse the html documentation locally:
+
+```bash
+make docs
+cd docs/_build/html
+python -m SimpleHTTPServer 8765
+# point web browser to http://localhost:8765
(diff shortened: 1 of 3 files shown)


HACKING.md
```diff
@@ -0,0 +1,30 @@
...
+```bash
+make docs
+```
+
+To browse the html documentation locally:
+
+```bash
+make docs
+cd docs/_build/html
+python -m SimpleHTTPServer 8765
```
johnsca: You don't actually need to run a server; just pointing to the static HTML files using a `file://` URL works fine.  I usually use `gnome-open docs/_build/html/index.html`

galgalesh: Thats a lot easier!

`gnome-open` isn't installed by default in 17.10, but `xdg-open` is, so I'll use that.

galgalesh:
Fixed the simple server stuff.

Wouldn't it be better to put further debugging information in the Juju docs themselves? So we don't have to maintain this info in two places?

johnsca:
Yep, it would.

Issue #137, opened 2017-10-30
Transaction nature of reactive framework needs to be explained better.


galgalesh:
From email thread: https://lists.ubuntu.com/archives/juju/2017-October/009548.html

The reactive framework resets state changes when a handler fails. This surprises charm authors. States are often used to ensure idempotency such as in the following example.

```python
@when_not("initialized")
def init():
    must_be_called_exactly_once()
    set_state("initialized")

@when("initialized")
@when_not("ready")
def get_ready():
    this_call_fails()
    set_state("ready")
```

After `this_call_fails` fails, the `initialized` state gets removed again, causing the `init` function to be called again during the next run. As a result of this behavior, all handlers must be idempotent and states cannot be used to achieve that.

The reason the charm state needs to be rolled back on hook failure is that the changes made to the Juju environment get rolled back on hook failure. However, since hooks are hidden from the charm author it is not clear when a hook starts and when it finishes, and this transactional nature isn't explained in the docs nor are the names of the commands self-explanatory.

ajkavanagh:
I tend to agree.  There are other ways to achieve a true "run this only once" (e.g. use your own sentinel file).  But there is a basic issue with the imperative nature of writing code in charms being in 'conflict' with juju's approach of retrying failed hooks.

Your example shows how very, very subtle bugs can be introduced into a charm.  Perhaps, in the docs, a more 'declarative' / 'goal-seeking' approach should be considered, where every shown function is written in a way that results in the same state regardless of how many times it is called (successfully)

Issue #117, opened 2017-06-30
Better charms.reactive.cli documentation


AdamIsrael:
The [charms.reactive.cli](https://charmsreactive.readthedocs.io/en/latest/charms.reactive.cli.html) documentation should include examples of how to use it, and references to any well-known charms that put it into practice.

It would also help to add some docs explaining what and why the cli api provide. There's some really cool bash tricks in there, particularly with regard to decorators, and we should talk about that so it's better understood.

Issue #22, opened 2015-11-09
only_once will be called multiple times


stub42:
The guard used to stop only_once decorated code from running more than once includes the code line number in its key. The method will be run again if it is moved around the source file, or into a different source file for that matter. Which is fine, as I can argue that running only_once decorated methods  once after every upgrade-charm hook is a good thing.

I think this should be explicit, rather than random, and the was_invoked/marked_invoked caches cleared at the beginning of an upgrade-charm hook. If people do need a only_once_really decorator, the developer would need to provide a key rather than a volatile one generated for them.

galgalesh:
I'd argue against making this explicit. I think the main use-case for `only-once` is a handler that isn't idempotent, therefore it should never be allowed to run, even during an upgrade-charm hook.

stub42:
We have the choice between:

- leaving things as they are and documenting the current behaviour

- having only_once reset on upgrade-charm so they are always rerun on upgrade-charm rather than sometimes rerun on upgrade-charm

- requiring or optionally allowing the developer to add an explicit key

- deprecating only_once.

galgalesh:
My vote goes to **deprecation**. The main use-case for `only-once` is a handler that isn't idempotent, such as formatting HDFS.

- Non-idempotent handlers are already an anti-pattern since a handler might run multiple times during debugging.
- The `only_once` functionality can be simulated by using states if you really need it.

Also, removing stuff just feels good :), removing complexity and stuff...

galgalesh:
@stub42 , @johnsca , 

How do we go about deprecating this? Is there any way for us to give warnings at `charm build` time? Such functionality will also be useful for the 2.0 switch.

ajkavanagh:
Over the years, I've written deprecate decorators of the form:

```python
import functools


def deprecate(warning, logger=None):

    def wrap(f):
        _call_count = 0

        @functools.wraps(f)
        def wrapped_f(*args, **kwargs):
            nonlocal _call_count
            if not _call_count:
                # log or print deprecation here
                print(warning)
                _call_count = 1
            return f(*args, **kwargs)
        return wrapped_f
    return wrap


@deprecate("Don't use this")
def deprecated_function():
    # do something
    print("deprecated_function")


@deprecate("This one either")
def f2():
    print("f2 away")


deprecated_function()
deprecated_function()
f2()
deprecated_function()
f2()
f2()
```

with the corresponding output:

    python3 test-deprecate.py
    Don't use this
    deprecated_function
[...]

PR #140, opened 2017-11-06
More doc fixes

galgalesh:
- Add CLI docs to API index
- Explain transactional nature of set_flag and similar functions
- DEPRECATED -> to sphinx `..:deprecated:` directive
- Hide deprecated methods in summary
- Override function signature to hide internal use only parameters

These are partial fixes for  #137, #22, #117

### DIFF
--- a/charms/reactive/decorators.py
+++ b/charms/reactive/decorators.py
@@ -210,6 +210,11 @@ def not_unless(*desired_flags):
 
 def only_once(action=None):
     """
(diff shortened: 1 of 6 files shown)


johnsca:
The test failure is very strange.  It's not a general snap issue, because the Juju snap does the bootstrap just fine.  It seems to be specific to the `juju-wait` snap, but I can't imagine what that snap would do differently.

PR #141, opened 2017-11-07
Fix test error from juju-wait snap

johnsca:
Work around seccomp error: https://bugs.launchpad.net/juju-wait/+bug/1730739


johnsca:
Didn't work.   Closing.

PR #148, opened 2017-12-12
Add deprecated placeholder for RelationBase.from_state

johnsca:
For backwards compatibility, in case anyone is using it.

Added info about all deprecations to the changelog to make them more visible.


johnsca:
Test failures are due to an issue with the charm store.  It seems to currently be failing for all charms.

PR #153, opened 2018-02-01
Separate departed units from joined in Endpoint

### LINKED ISSUE #151: joined flag only removed in the hook after the relation-broken hook
The docs say the following

> the `joined` flag will be automatically cleared when there are no remote units remaining on any relation, but neither of the other flags will be automatically cleared.

I read this as "the `joined` flag will be removed **when the last unit of the last relationship is `departed`**". So for an endpoint with one relation and one remote unit, the `joined` flag will be removed at the same time that the `departed` flag is set.

This is however not the case: the `joined` flag is only removed when during the hook _after_ the  `relation-broken` hook because that is the first time when Juju says that there are no relations/units connected anymore.

Look at this output. The desired behavior is that the `joined` flag is removed in the first `departed` hook, not in the `update-status` hook.

``` bash
# Departed hook
2018-01-18 13:16:30 INFO juju-log kubernetes-request:94: Reactive main running for hook kubernetes-request-relation-departed
2018-01-18 13:16:31 INFO juju-log kubernetes-request:94: Invoking reactive handler: reactive/kubernetes-resource.py:11:print_flags
2018-01-18 13:16:31 INFO juju-log kubernetes-request:94: [
  'endpoint.kubernetes-request.changed', 
  'endpoint.kubernetes-request.changed.egress-subnets', 
  'endpoint.kubernetes-request.changed.ingress-address', 
  'endpoint.kubernetes-request.changed.private-address', 
  'endpoint.kubernetes-request.departed', 
  'endpoint.kubernetes-request.joined', 
  'kubernetes.connected']

# Broken hook
2018-01-18 13:16:32 INFO juju-log kubernetes-request:94: Reactive main running for hook kubernetes-request-relation-broken
2018-01-18 13:16:32 INFO juju-log kubernetes-request:94: Invoking reactive handler: reactive/kubernetes-resource.py:11:print_flags
2018-01-18 13:16:32 INFO juju-log kubernetes-request:94: [
  'endpoint.kubernetes-request.changed', 
  'endpoint.kubernetes-request.changed.egress-subnets', 
  'endpoint.kubernetes-request.changed.ingress-address', 
  'endpoint.kubernetes-request.changed.private-address', 
  'endpoint.kubernetes-request.departed', 
  'endpoint.kubernetes-request.joined', 
  'kubernetes.connected']

# Update status hook
2018-01-18 13:21:27 INFO juju-log Reactive main running for hook update-status
2018-01-18 13:21:27 INFO juju-log Invoking reactive handler: reactive/kubernetes-resource.py:16:k8s_resource_install
2018-01-18 13:21:27 INFO juju-log NO DEPLOYER RELATION
2018-01-18 13:21:27 INFO juju-log Invoking reactive handler: reactive/kubernetes-resource.py:11:print_flags
2018-01-18 13:21:28 INFO juju-log [
  'endpoint.kubernetes-request.changed',
  'endpoint.kubernetes-request.changed.egress-subnets',
  'endpoint.kubernetes-request.changed.ingress-address',
  'endpoint.kubernetes-request.changed.private-address',
  'endpoint.kubernetes-request.departed']
```

This behavior requires a developer to know the intricacies of when Juju stops saying there is a relationship.

I propose we remove the `joined` flag, during the `departed` hook of the last unit of the last relationship.

1. It's more intuitive,
2. it's how it's explained in the docs,
3. and the current behavior serves no purpose, afaik.

What do you think, @johnsca ?

### LINKED ISSUE #152: Improvement endpoint docs
@galgalesh It would be nice if the docs concerning the new endpoint implementation could highlight some major changes that affect the behaviour of flags. 

For example:

> These are mainly intended to be used by the interface layer itself, but can be documented as part of the official API that the interface layer exposes. Additionally, the joined flag will be automatically cleared when there are no remote units remaining on any relation, but neither of the other flags will be automatically cleared. 

This is a big difference and is easy to miss in the wall of text. A notification box or some kind of warning would be helpful.

johnsca:
This fixes #151 by adding a `departed_units` collection which doesn't count toward joined units when determining if the endpoint is still considered joined.

**Edit:** Also fixes #152

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,21 +4,20 @@ language: python
 python:
   - "3.5"
 env:
-  - SNAP_CMD="sudo snap install juju --classic --stable"
-  - SNAP_CMD="sudo snap install juju --classic --edge"
+  - SNAP_CHANNEL="stable"
+  - SNAP_CHANNEL="edge"
 before_install:
-  - sudo add-apt-repository ppa:ubuntu-lxc/lxd-stable -y
-  - sudo apt-get update -q
-  - sudo apt-get install lxd snapd pwgen -y
-  - sudo usermod -a -G lxd $USER
-  - sudo service lxd start || true
-  - sudo lxd init --auto
+  - sudo apt-get install snapd pwgen -y
+  - sudo apt-get remove -qy lxd lxd-client
 install:
   - pip install tox-travis
-  - (eval "$SNAP_CMD")
-  - sudo snap install charm --edge
-  - sudo snap install --classic juju-wait
+  - sudo snap install lxd --$SNAP_CHANNEL
+  - sudo snap install juju --classic --$SNAP_CHANNEL
+  - sudo snap install charm --$SNAP_CHANNEL
+  - sudo snap install --classic juju-wait --$SNAP_CHANNEL
 before_script:
+  - sudo addgroup lxd || true
+  - sudo usermod -a -G lxd $USER
   - sudo -E sudo -u $USER -E bash -c "/snap/bin/juju bootstrap localhost test"
 script:
   - tox -e lint,py35
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ ftest: lint
 	.tox/py3/bin/nosetests --attr '!slow' --nologcapture tests/
 
 docs: lint
-	.tox/py3/bin/pip install sphinx sphinx_rtd_theme recommonmark
+	.tox/py3/bin/pip install 'sphinx>=1.7b2' sphinx_rtd_theme recommonmark
 	(cd docs; make html SPHINXBUILD=../.tox/py3/bin/sphinx-build)
 	cd docs/_build/html && zip -r ../docs.zip *
 .PHONY: docs
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -164,7 +164,7 @@ class Endpoint(RelationFactory):
         Collection of :class:`Relation` instances that were previously
         established for this :class:`Endpoint` but which are no longer.
 
-        Like :prop:`.relations`, this is a :class:`KeyList`.  However, it is
+        Like :attr:`relations`, this is a :class:`KeyList`.  However, it is
         also both persistent and mutable.  Once a broken relation has been
         handled, however is appropriate for the charm, it can be deleted from
         this collection.  A broken relation will eventually be cleaned up
@@ -279,7 +279,7 @@ class Endpoint(RelationFactory):
         be kept until they are explicitly removed, to allow for reasonable
         cleanup of units that have left.
 
-        Once a unit is departed, it will no longer show up in :prop:`.all_units`.
+        Once a unit is departed, it will no longer show up in :attr:`all_units`.
         Note that units are considered departed as soon as the departed hook
         is entered, which differs slightly from how the Juju primitives behave
         (departing units are still returned from ``related-units`` until after
@@ -376,7 +376,7 @@ class Relation:
         be kept until they are explicitly removed to allow for reasonable
         cleanup of units that have left.
 
-        Once a unit is departed, it will no longer show up in :prop:`.units`.
+        Once a unit is departed, it will no longer show up in :attr:`units`.
         Note that units are considered departed as soon as the departed hook
         is entered, which differs slightly from how the Juju primitives behave
         (departing units are still returned from ``related-units`` until after
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -16,7 +16,6 @@
 
 import json
 from collections import UserDict
-from functools import partial
 from itertools import chain
 
 from charmhelpers.core import hookenv, unitdata
@@ -131,8 +130,8 @@ class Endpoint(RelationFactory):
         self._endpoint_name = endpoint_name
         self._relations = KeyList(map(Relation, relation_ids or []),
                                   key_attr='relation_id')
-        self._broken_relations = None
-        self._all_units = None
+        self._all_joined_units = None
+        self._all_departed_units = None
 
     @property
     def endpoint_name(self):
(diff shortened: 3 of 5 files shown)


johnsca:
Travis failure on edge will be fixed by juju/charm-tools#387

galgalesh:
Awesome! I looked at this together with @xannz and the flags work exactly as expected.

I see a possible issue with `departed_units`: one hook after the relation-broken hook, `self.relations` won't contain that relation anymore, which means it _does_ matter in which hook you handle a departed unit in the case that it's the last unit of the relation.

I'm not sure this is an issue, but it might be unexpected. I don't want to add a `self.broken_relations` because that re-introduces the concept of a relation that's not broken but that has no units, and I think that only makes things more complicated.

So what do you think about also caching all relations? Charm authors can still use `relation.joined` to see if the relation has any units..?

johnsca:
Hrm.  I had missed that.  If we're going to cache relations, I'd prefer to do it in a separate collection so that in the common case, the user can just iterate over `endpoint.relations` and `rel.units` and always only see the things that are actually joined.  We could handle the corner case the same as we now do with departing units: in reactive, a relation is "broken" as soon as its last relation is departing (i.e., its `rel.units` collection is empty).

Also note that I didn't end up caching all units, just the ones as they departed.  I determined this ended up being easier because that way I didn't have to deal with ensuring that the relation data cache got updated at a reasonable time, but I actually wonder if it might be better to go ahead and add the caching for everything since that would also allow us to 1) avoid calls to the agent which have been a source of extra, unexpected network traffic to the controller, 2) enforce consistency about when changes to charm state are visible (there's currently some ambiguity with, e.g., config data and possibly relation data because the data that's visible can be different prior to the relevant `-changed` hook triggering, though that's less of an issue when doing reactive "properly").  We technically already have a cache of the config data (in `hookenv.config`), but it's updated on every hook (which is where we first noticed the extra network traffic, always requesting all config during `update-status`); we could change that to only update on `config-changed` and make cached relation data only update on `relation-x-changed`.  Thoughts?

johnsca:
I had also considered creating an `Endpoint.all_departed_units` view but figured I'd leave that for down the road.  But another option would be to have departed units be considered "orphaned" from their relation, and so only live on an `Endpoint` level collection.  If we have a `broken_relations` collection, it means you'd have to iterate over both `endpoint.relations.departed_units` and `endpoint.broken_relations.departed_units` to get all of the departed units, and you wouldn't be able to tell which collection the currently leaving unit was in.

johnsca:
[Better handling of departing units / broken relations in Endpoints](https://trello.com/c/eLDEcS9y/175-better-handling-of-departing-units-broken-relations-in-endpoints)

galgalesh:
> I actually wonder if it might be better to go ahead and add the caching for everything since that would also allow us to 1) avoid calls to the agent which have been a source of extra, unexpected network traffic to the controller, 2) enforce consistency about when changes to charm state are visible (there's currently some ambiguity with, e.g., config data and possibly relation data because the data that's visible can be different prior to the relevant -changed hook triggering, though that's less of an issue when doing reactive "properly").

I don't think `2` is a real issue for us because a reactive charm shouldn't care about which hook something happens in. I think it's actually hard to write a handler that is hook-specific using the latest charms.reactive.

The current behavior also has its advantages, a deployment can settle sooner, because you avoid charms re-configuring the application using outdated info: more hooks will be no-ops.

> If we have a broken_relations collection, it means you'd have to iterate over both endpoint.relations.departed_units and endpoint.broken_relations.departed_units to get all of the departed units, and you wouldn't be able to know ahead of time which collection the currently leaving unit was in. 

I agree, `broken_relations` doesn't seem the right solution for this.


>  But another option would be to have departed units be considered "orphaned" from their relation, and so only live on an Endpoint level collection.

This seems like a good option too, especially because it allows us to do the following:

> We could handle the corner case the same as we now do with departing units: in reactive, a relation is "broken" as soon as its last relation is departing (i.e., its rel.units collection is empty).

johnsca:
I ended up doing both; there's now an `all_departed_units` collection on the `Endpoint` as well as a `broken_relations` collection.  Relations are moved to the `broken_relations` collection as soon as their last unit starts the `-relation-departed` hook.

I also fixed the issue where not all of the classes were being included in the API reference docs due to the module's `__all__` list.  To do this, I had to bring in the beta version of Sphinx (1.7b2) to get the new `ignore-module-all` option.  With this and @galgalesh's earlier doc improvements in this PR, this PR also fixes #152.

galgalesh:
Just want to say that I really like this idea of a writeable collection of `all_departed_units` and `broken_relations`. I think this is really useful!

That said, I'm not convinced that it's good to keep `relation.departed_units`. We risk Charmers writing code like the following:

```python
for rel in self.relations:
    for un in rel.departed_units:
        # handle departed unit
        # remove departed unit from collection
```

Which will work for every departed unit except the last one. The only correct way to use `departed_units` is the following:

```python
for rel in self.relations:
    for un in rel.departed_units:
        # handle departed unit
        # remove departed unit from collection
for rel in self.broken_relations:
    for un in rel.departed_units:
        # handle departed unit
        # remove departed unit from collection
```

Which is a lot less cleaner than

```python
for un in self.all_departed_units:
    # handle departed unit
    # remove departed unit from collection
```

Moreover, if a charmer really wants to iterate departed units of a single relation, he can always do so by filtering `self.departed_units`. This is a little bit more complex, but I don't think it's a common use-case so I think that's a good tradeoff for preventing a charmer from shooting itself in the foot.

**Smal note**

Just an idea that popped into my head: what do you think about making an alias `self.joined_relations` for `self.relations` and pointing to `joined_relations` in the docs? We could do the same with `relation.joined_units`. This makes it explicit, useful for charmers who don't read the docs.

johnsca:
If we remove `relation.departed_units`, then we lose the symmetry for `relation.joined_units`.  Though, maybe we could shorten the `Endpoint` collections to `endpoint.joined_units` and `endpoint.departed_units`?

I'm also not sure if `endpoint.broken_relations` is even useful if we only have `endpoint.all_departed_units`.  Maybe we should drop that too?

galgalesh:
> Though, maybe we could shorten the Endpoint collections to endpoint.joined_units and endpoint.departed_units and then have the Relation collection match that (relation.joined_units)?

I like the difference between `endpoint.all_units` and `relation.units`, because it helps the charmer know that these are two different things. It's a small thing but it might help during debugging, because the difference jumps out more.

> I'm also not sure if endpoint.broken_relations is even useful if we only have endpoint.all_departed_units. Maybe we should drop that too?

I agree.

johnsca:
So, `endpoint.all_joined_units`, `endpoint.all_departed_units`, and `relation.joined_units`?

galgalesh:
yes :+1:

johnsca:
Updated per discussion.

ktsakalozos:
Although I haven't followed the discussion and the evolution of this code, I am +1 to this work. It will save us from a few headaches. Many thanks.

galgalesh:
\o/

PR #163, opened 2018-02-28
Fix setdefault on Endpoint data collections

johnsca:
`UserDict`'s implementation of `setdefault` doesn't work with the `None` default value in our data collections, leading to `setdefault` always returning `None` and never updating the collection.

### DIFF
--- a/tests/test_endpoints.py
+++ b/tests/test_endpoints.py
@@ -383,6 +383,10 @@ class TestEndpoint(unittest.TestCase):
         rel.to_publish.update({'key': {'new': 'new'}})
         self.assertEqual(rel.to_publish_raw, {'key': '{"new": "new"}'})
 
+        assert isinstance(rel.to_publish.get('foo', {}), dict)
+        assert isinstance(rel.to_publish.setdefault('foo', {}), dict)
+        assert isinstance(rel.to_publish['foo'], dict)
+
     def test_handlers(self):
         Handler._HANDLERS = {k: h for k, h in Handler._HANDLERS.items()
                              if hasattr(h, '_action') and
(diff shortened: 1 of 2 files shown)


tests/test_endpoints.py
```diff
@@ -383,6 +383,10 @@ def test_to_publish(self):
         rel.to_publish.update({'key': {'new': 'new'}})
         self.assertEqual(rel.to_publish_raw, {'key': '{"new": "new"}'})
 
+        assert isinstance(rel.to_publish.get('foo', {}), dict)
+        assert isinstance(rel.to_publish.setdefault('foo', {}), dict)
```
stub42: If you change this to setdefault('foo', 'other'), then you are confirming that setdefault is not resetting the dict.

You don't have tests that:
1) Use to_publish.setdefault to set a missing value
2) Use to_publish['foo'] to change a value unconditionally
3) Use to_publish.setdefault to not change an existing value
4) Use del to_publish['foo'] to remove an existing value 
5) Use del to_publish['foo'] to fail with a KeyError if the value is missing (or should this be a noop?)
6) Use to_publish['missing'] to fail with a KeyError if the value is missing (or should it return None or the empty string?)

(the last two are subtle but important, as we may need to match relation-set's behaviour rather than following Python conventions. to_publish_raw might have some oddities with to_publish_raw['foo'] = None vs to_publish_raw['foo'] = '' , but we can get away with weirdness on the _raw interface)

johnsca: To clarify, the `to_publish['missing'] is None` is intended to match the behavior of `relation-get` which returns `null` when asked to fetch a relation key that doesn't exist, rather than raising an error.  But I'm still not clear if is less confusing to replicate that in Python or if it would be better to replicate the Python behavior.

stub42: to_publish['foo'] = None should set the relation attribute to the string 'null' because it gets JSON encoded.
Thus, to_publish['missing'] should raise a KeyError, while to_publish['foo'] would return None.

to_publish_raw would be where we want to match relation-get behaviour.

PR #164, opened 2018-03-08
Remove state when its value is "None"

tytus-kurek:
This patchset implements a logic to remove the state when its value
is set to "None".

Partial-Bug: 1721580

### DIFF
--- a/charms/reactive/relations.py
+++ b/charms/reactive/relations.py
@@ -702,6 +702,7 @@ class Conversation(object):
         state = state.format(relation_name=self.relation_name)
         value = _get_flag_value(state)
         if not value:
+            clear_flag(state)
             return
         if self.key in value['conversations']:
             value['conversations'].remove(self.key)
--- a/tests/test_relations.py
+++ b/tests/test_relations.py
@@ -464,16 +464,15 @@ class TestConversation(unittest.TestCase):
         conv.remove_state('{relation_name}.bar')
         get_flag_value.assert_called_once_with('rel.bar')
         assert not set_flag.called
-        assert not clear_flag.called
+        assert clear_flag.called
 
         conv.remove_state('{relation_name}.bar')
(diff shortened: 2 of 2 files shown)


galgalesh:
@tytus-kurek 

Thanks for the PR, although I'm not sure I understand why this is needed. Can you explain why the state/flag needs to be removed? Give a use-case where this is important?

Can you also give link to the bug report? I'm not sure which bugtracker you're referring to. Also please explain why this PR will (partially) fix that issue.

The fact that you have to revert a test worries me; it means that we explicitly intended these states not to be removed, so there might be valid use-cases for this behavior.


Thanks!

ajkavanagh:
Having reviewed the bug (https://bugs.launchpad.net/charm-interface-rabbitmq/+bug/1721580) I reckon this isn't an appropriate fix.  The interface should remove the state and not set it to None.  The code in the PR also just checks for False-y values, and thus a flag set to 'False' would be removed; which I suspect would be surprising behaviour.  I don't think this is an appropriate code fix.

charms/reactive/relations.py
```diff
@@ -702,6 +702,7 @@ def remove_state(self, state):
         state = state.format(relation_name=self.relation_name)
         value = _get_flag_value(state)
         if not value:
+            clear_flag(state)
```
ajkavanagh: If `value` is false-y then the state will be cleared; this is not what the PR is titled with, which suggests None.

tytus-kurek:
@ajkavanagh I am happy to update the fix.

"The interface should remove the state"

This has been already proposed at https://review.openstack.org/#/c/550770/

"and not set it to None"

At the moment this state is set by the designate charm, not by the interface. Could you please guide me what value should be assigned to the state if not "None"?

ajkavanagh:
okay, I reckon we should close this PR without merging.  The actual fix will be in the charm layer.

PR #165, opened 2018-03-15
Fix documentation of unit_name and application_name on RelatedUnit

johnsca:
Make them properties so that they're immutable and show up in the docs.


galgalesh:
LGTM :+1:

PR #167, opened 2018-04-02
Fix hard-coded version in docs config

johnsca:


johnsca:
Test failure is a failure to bootstrap on juju edge.  Stable passes, so merging this.

PR #168, opened 2018-04-11
Rename Endpoint.joined to Endpoint.is_joined

johnsca:
It came up that an interface layer could easily want to name a handler `joined` which ends up conflicting with the property and breaking the management of the `endpoint.{endpoint_name}.joined` flag causing it to always be set, even when no related units are joined.

The `is_joined` name is more descriptive to indicate that it's a boolean flag, anyway.

### DIFF
--- a/tests/test_endpoints.py
+++ b/tests/test_endpoints.py
@@ -164,12 +164,14 @@ class TestEndpoint(unittest.TestCase):
         assert Endpoint.from_name('test-endpoint') is not None
         assert Endpoint.from_name('test-endpoint').endpoint_name == 'test-endpoint'
         assert Endpoint.from_name('test-endpoint').is_joined
+        assert Endpoint.from_name('test-endpoint').joined  # deprecated
         assert is_flag_set('endpoint.test-endpoint.joined')
         assert is_flag_set('endpoint.test-endpoint.changed')
         assert is_flag_set('endpoint.test-endpoint.changed.foo')
         assert Endpoint.from_name('test-endpoint2') is not None
         assert Endpoint.from_name('test-endpoint2').endpoint_name == 'test-endpoint2'
         assert not Endpoint.from_name('test-endpoint2').is_joined
+        assert not Endpoint.from_name('test-endpoint2').joined  # deprecated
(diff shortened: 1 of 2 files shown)


johnsca:
For example, I hit this myself in juju-solutions/interface-http#10

stub42:
I was wondering the other day if adding handlers on the Endpoint class should be encouraged at all. The Endpoint implementation provides an API for charms to use, and the handlers are never going to be part of that API. Things are cleaner if the handlers are declared as normal functions elsewhere in the .py file, retrieving the Endpoint using reactive.endpoint_from_flag like a charm does. I don't think any functionality is lost.

johnsca:
We would still need to subclass `Endpoint` to add accessors and methods for the handlers to call, which could still conflict, but certainly the handlers don't need to be methods.  I wonder if we could make it so that it uses has-a instead of is-a somehow to scope the framework provided API down?  Something like `self.endpoint.is_joined` and `self.endpoint.relations[0].joined_units`, though there's probably a better word than "endpoint" again.

johnsca:
@stub42 I just tried to implement an interface layer using your idea of having interface layer handlers not be methods on the class, and ran into an issue with that approach.  Specifically, the interface layer can't know what the endpoint name is, so it can't use `endpoint_from_flag`.  Charms don't have this issue because they know the endpoint name a priori.  You can use the old positional param approach, but I'd like to move away from that because it's harder to reason about.  Any suggestions on how to get around that?

stub42:
@johnsca Oh, right. The function needs state, and the only way of passing the state is as a parameter. I'd stick with what we have. The Endpoint can always prefix its handler's with _ to declare them private if the author cares.

tests/test_endpoints.py
```diff
@@ -163,13 +163,13 @@ def test_startup(self):
         Endpoint._startup()
         assert Endpoint.from_name('test-endpoint') is not None
         assert Endpoint.from_name('test-endpoint').endpoint_name == 'test-endpoint'
-        assert Endpoint.from_name('test-endpoint').joined
```
stub42: I wouldn't remove the .joined checks, since the method needs to remain for backwards compatibility. Just add the new checks with is_joined.

PR #176, opened 2018-05-02
Add missed deprecated backwards compatibility import alias for State

johnsca:


johnsca:
Travis failure is Juju edge failing to bootstrap.  Juju stable passed, so going to merge anyway,

Issue #180, opened 2018-06-04
Endpoint JSON fails with Trusty + Python 3.4


stub42:
charms.reactive endpoints is not working with the Python 3.4 json library.
It is attempting to access the json.JSONDecodeError symbol which does not exist with Python 3.4

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-dse-11/.venv/lib/python3.4/site-packages/charms/reactive/__init__.py", line 71, in main
    hookenv._run_atstart()
  File "/var/lib/juju/agents/unit-dse-11/.venv/lib/python3.4/site-packages/charmhelpers/core/hookenv.py", line 1112, in _run_atstart
    callback(*args, **kwargs)
  File "/var/lib/juju/agents/unit-dse-11/.venv/lib/python3.4/site-packages/charms/reactive/endpoints.py", line 131, in _startup
    endpoint._manage_flags()
  File "/var/lib/juju/agents/unit-dse-11/.venv/lib/python3.4/site-packages/charms/reactive/endpoints.py", line 226, in _manage_flags
    for key, value in unit.received.items():
  File "/var/lib/juju/agents/unit-dse-11/.venv/lib/python3.4/_collections_abc.py", line 498, in __iter__
    yield (key, self._mapping[key])
  File "/var/lib/juju/agents/unit-dse-11/.venv/lib/python3.4/site-packages/charms/reactive/endpoints.py", line 832, in __getitem__
    except json.JSONDecodeError:
AttributeError: 'module' object has no attribute 'JSONDecodeError'
```

johnsca:
Fixed with #181

PR #181, opened 2018-06-04
Catch Exception rather than json.JSONDecodeError for Python 3.4

stub42:
Catching Exception should be good enough, as the only call being made
is json.loads() and it won't catch system exceptions such as memory
issues. Still, we should tighten it back up once we can drop
Python 3.4 support with the Ubuntu 14.04 EOL.

Addresses #180


johnsca:
Test failure is unrelated to this change, but I'm a little concerned that `charm pull-source` is broken on edge of `charm`.

PR #183, opened 2018-06-25
Add upgrade.series.{pre,post} flags

johnsca:
This also adds documentation for all existing automatically managed flags.

### DIFF
--- a/docs/managed-flags.rst
+++ b/docs/managed-flags.rst
@@ -1,0 +1,88 @@
+.. _automatic-flags:
+
+Automatic Flags
+===============
+
+The reactive framework will automatically set some flags for your charm,
+based on lifecycle events from Juju.  These flags can inform your charm
+of things such as upgrades, config changes, relation activity, etc.
+
+With a few exceptions, noted below, these flags will be set by the framework
+but it is up to your charm to clear them, if need be.  To avoid conflicts
+between layers, it is recommended that only the top-level charm layer (or
+interface layer, in the case of endpoint flags) use any of the automatic flags
+directly; any base layer should instead use
+:func:`~charms.reactive.flags.register_trigger` to "wrap" the automatic flag
+with a layer-specific flag that can be safely used within that layer.
+
+The flags that are set by the framework are:
+
++----------------------------------------------+------------------------------------------------------------+
+| ``upgrade.series.pre``                       | This is set when the ``pre-series-upgrade`` hook is        |
+|                                              | fired, indicating that any application services should     |
+|                                              | be stopped and disabled so that they do not start on       |
+|                                              | reboot, so that the operator can perform a OS series       |
+|                                              | upgrade.                                                   |
++----------------------------------------------+------------------------------------------------------------+
+| ``upgrade.series.post``                      | This is set when the ``post-series-upgrade`` hook is       |
+|                                              | fired, indicating that the operator has completed the      |
+|                                              | series upgrade and any application migrations should be    |
+|                                              | performed and any services should be resumed and           |
+|                                              | re-enabled.                                                |
++----------------------------------------------+------------------------------------------------------------+
+| ``config.changed``                           | This is set when any config option has changed. [1]_       |
++----------------------------------------------+------------------------------------------------------------+
+| ``config.changed.{option_name}``             | This is set for each config option that has changed. [1]_  |
++----------------------------------------------+------------------------------------------------------------+
+| ``config.set.{option_name}``                 | This is set for each config option whose value is not      |
+|                                              | ``None``, ``False``, or an empty string. [1]_              |
++----------------------------------------------+------------------------------------------------------------+
+| ``config.default.{option_name}``             | This is set for each config option whose value was         |
+|                                              | changed from its default. [1]_                             |
++----------------------------------------------+------------------------------------------------------------+
+| ``leadership.is_leader``                     | This is set when the unit is the leader. The unit will     |
+|                                              | remain the leader for the remainder of the hook, but       |
+|                                              | may not be leader in future hooks. [2]_                    |
++----------------------------------------------+------------------------------------------------------------+
+| ``leadership.changed``                       | This is set when any leadership setting has changed. [2]_  |
++----------------------------------------------+------------------------------------------------------------+
+| ``leadership.changed.{setting_name}``        | This is set for each leadership setting that has           |
+|                                              | changed. [2]_                                              |
++----------------------------------------------+------------------------------------------------------------+
+| ``leadership.set.{setting_name}``            | This is set for each leadership setting that has been      |
+|                                              | to set to a value other than ``None``. [2]_                |
++----------------------------------------------+------------------------------------------------------------+
+| ``endpoint.{endpoint_name}.joined``          | This is set when a relation is joined on an endpoint. [3]_ |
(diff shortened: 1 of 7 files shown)


johnsca:
One comment from the larger PR (#182) was why not just use the hook name as the flag.  I still didn't go that route here because I think that the pattern of the flag names is fairly established with the existing flags, where the flag name goes from the more general prefix to more specific to the right.  I did change the flags slightly, however, so that they are just the hook name components reversed.  Hopefully that, along with the additional documentation, is enough to allay the concerns raised.

stub42:
Another issue I just realized with requiring the charm to clear the flags is how does a charm start using them safely?

I imagine this is how you hope to see them used:
```python
@when('upgrade.series.pre')
def prepare_dist_upgrade():
    host.service_stop('foo')
    apt.purge('foo')
    reactive.clear_flag('upgrade.series.pre')
```

The problem is, what if there had previously been a distro upgrade on this unit? The flag will have already been set and not cleared, because the charm didn't care before. When the new code lands to support upgrades better, the handler will run and maybe harmful things happen because a distro upgrade isn't actually happening; the flags were set by charms.reactive, and clearing them deferred to a charm that didn't care. And when the charm starts caring, it lacks information to handle the flag lifecycle correctly.

You can't resolve the situation n a `@hook('upgrade-charm')` or `@when('hook.upgrade-charm')` handler, as handler does not know what state the system was in when the flag got set, so does not know if it should be cleared. It can either leave the flag set and hope the handlers are idempotent, or clear the flags and risk skipping handlers that should be run. Triggers, however, do provide an option, provided charms.reactive always clears the flag and resets it, causing triggers to fire even if the flag was already set. Which is adding more confusion, because you now have some flags that fire their triggers even when the flag does not change state, which is not the default behavior.

I think this applies to all reactive managed flags to some extent, including the most common like config.changed. It pretty much mandates that handlers using them for guards must be idempotent, or requires using triggers in all cases rather than just in base layers.

```python
register_trigger(when='upgrade.series.pre', set_flag='my.upgrade.series.pre')

@when('my.upgrade.series.pre')
def prepare_dist_upgrade():
    host.service_stop('foo')
    apt.purge('foo')
    reactive.clear_flag('my.upgrade.series.pre')
```

I'm not feeling that this direction change is for the better. I think things are getting worse for charm authors. and it will be harder to write correct charms than it is now. There is more to understand, not less. If we must avoid 'set for the duration of the hook', then I think the problems can be avoided by setting and immediately clearing the flag. Charm authors would be required to use register_trigger, which would do the right thing. I think this is preferable to allowing charm authors to use alternative mechanisms, which are potentially foot guns. Best practice would become 'always use register_trigger', and we end up in the same situation but with more confusing documentation and needing to maintain the behavior for backwards compatibility.

I also personally don't believe that 'set for the duration of the hook' is actually a problem, and that charm authors are quite able to understand and reason about what this means. I personally think the best way forward is to keep the flag lifecycles as they are, where they are cleared automatically at the end of the hook. Documented best practice would be to use register_trigger, because the pattern avoids the various race conditions entirely. We end up with simpler and more correct charms using a single, no matter if reading code in a charm layer or a base layer. And we never have to worry about changing how the charms.reactive flag lifecycles are maintained and backwards compatibility, because it doesn't matter when following best practice.

I think we would be better off just documenting how to use triggers, and maybe coming up with easier ways to use them, making the flag lifecycle changes moot.

johnsca:
> I also personally don't believe that 'set for the duration of the hook' is actually a problem

The problem that I have with it in general is that the end of the hook context doesn't have semantic meaning relative to the information the flag is conveying.  For instance, `endpoint.{endpoint_name}.joined` doesn't stop being true just because the `relation-foo-joined` hook is done, so it doesn't make sense to remove it.  But in the same vein, neither does it become no longer true that the config had changed at the end of whatever hook the charm happened to first see the new values.

In the case of the `.joined` flag, there is a clear semantic point at which the framework can say that it's no longer true and it makes sense to remove it.  However, with things like `.changed`, it's impossible for the framework to make any reasonable statement about when it's been handled, so the most reasonable thing to do there is to just leave it set and leave it up to the charm to figure it out.  We can provide guidelines as to how and when to clear the flags, or provide mechanisms such that it's no longer necessary to ever clear them (triggers and / or a `force_flag`), but at the end of the day, it's really only the charm author that can sort it out.

All of that said, in the case of these particular hooks, as I thought about them more and more in coming up with this reply, I realized that we really shouldn't be treating them as independent events, but rather more like the `-joined` and `-departed` pair of hooks, in that they signal the start and end of something.  So I think I will change them to be a single flag, such as `upgrade.series.in-progress` which is set at the start of `pre-series-upgrade` and removed at the start of `post-series-upgrade`.

galgalesh:
> The problem is, what if there had previously been a distro upgrade on this unit? The flag will have already been set and not cleared, because the charm didn't care before. When the new code lands to support upgrades better, the handler will run and maybe harmful things happen because a distro upgrade isn't actually happening.

@johnsca's solutions to this seems like a good one to me. You could also solve this with a `@when_not('upgrade.series.post')` but then the charm author is basically manually simulating the behavior of @johnsca's solution so his seems better to me.

> For instance, endpoint.{endpoint_name}.joined doesn't stop being true just because the relation-foo-joined hook is done, so it doesn't make sense to remove it. But in the same vein, neither does it become no longer true that the config had changed at the end of whatever hook the charm happened to first see the new values.

:+1:

> I think we would be better off just documenting how to use triggers, and maybe coming up with easier ways to use them, making the flag lifecycle changes moot.

I like this solution for fixing the issues with the behavior of the `config.changed` flags although that should become a separate conversation. Maybe we should open an issue for further discussion on the `config.changed` flags and easier use of triggers?

docs/managed-flags.rst
```diff
@@ -0,0 +1,86 @@
...
+between layers, it is recommended that only the top-level charm layer (or
+interface layer, in the case of endpoint flags) use any of the automatic flags
+directly; any base layer should instead use
+:func:`~charms.reactive.flags.register_trigger` to "wrap" the automatic flag
+with a layer-specific flag that can be safely used within that layer.
+
+The flags that are set by the framework are:
+
++----------------------------------------------+------------------------------------------------------------+
+| ``upgrade.series.in-progress``               | This is set when the ``pre-series-upgrade`` hook is        |
```
galgalesh: Since these are new hooks, most charmers will probably first use them as a flag, not as a hook. Thus, the explanation of "this is set when x hook is fired" is mostly useless and requires the charmer to now look up this hook in the Juju docs in order to get more info. We should just copy the explanation of the hook in the Juju docs to here. Is there any draft for that explanation?

Also, I think it's best for the Juju docs to first explain the behavior of the flag (with code examples of a typical use case) and then mention the hook as a "low-level" way to do series migration for non-reactive, legacy charms.

galgalesh: Since these are new hooks, most charmers will probably first use them as a flag, not as a hook. Thus, the explanation of "this is set when x hook is fired" is mostly useless and requires the charmer to now look up this hook in the Juju docs in order to get more info. We should just copy the explanation of the hook in the Juju docs to here. Is there any draft for that explanation?

Also, I think it's best for the Juju docs to first explain the behavior of the flag (with code examples of a typical use case) and then mention the hook as a "low-level" way to do series migration for non-reactive, legacy charms.

PR #188, opened 2018-08-28
Add non-destructive version of data_changed

johnsca:
Having `data_changed` always update the hash can be problematic, and it is often useful to be able to check for an update without indicating that the hash should be updated.

### DIFF
--- a/charms/reactive/helpers.py
+++ b/charms/reactive/helpers.py
@@ -129,6 +129,17 @@ def mark_invoked(invocation_id):
     unitdata.kv().set('reactive.invoked.%s' % invocation_id, True)
 
 
+def _data_changed(data_id, data, hash_type, update):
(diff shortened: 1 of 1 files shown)


johnsca:
Not sure what's going on with the edge test; it's failing to bootstrap due to a lxd permission issue.  Worked fine on Juju stable.

johnsca:
lol.  If you think `has_data_changed` is more natural, I'm not averse to changing it.  `is_` is just my go-to prefix for boolean checks unless it's testing membership, but I can see your point on "has the data changed" being more natural when actually saying it.

stub42:
I think is_data_changed is correct here. In the case of a period where data changed and was then reverted, 'is_data_changed' should unambiguously return false, but 'has_data_changed' could return true.

PR #191, opened 2018-09-18
Flag and handler trace

stub42:
Working out which handlers will fire and when seems difficult for many developers starting out with charms.reactive. I think an option to log flag changes and the effects they trigger will help them understand the logic and track down bugs and race conditions.

For an example of what this ends up looking like, see https://pastebin.ubuntu.com/p/GHh8qyG7Yy/

I imagine a few ways that a developer could enable this optional behavior:
 * Creating a file `reactive/trace.py` containing `from charms.reactive import trace; trace.install_tracer(trace.LogTracer())`
 * Including a layer that does the above
 * A layer.yaml option that the base layer uses to turn on the logging.

It may even be worth just leaving it on (rather than consider this a developer option) if the performance hit and log verbosity is not considered a problem. Having this information in logs attached to bug reports could be really useful.

### DIFF
--- a/charms/reactive/bus.py
+++ b/charms/reactive/bus.py
@@ -379,10 +379,12 @@ def dispatch(restricted=False):
     _invoke(hook_handlers)
 
     unitdata.kv().set('reactive.dispatch.phase', 'other')
-    tracer().start_dispatch_loop()
     for i in range(100):
         FlagWatch.iteration(i)
         other_handlers = _test(Handler.get_handlers())
+        if i == 0:
+            tracer().start_dispatch_phase('other', other_handlers)
+        tracer().start_dispatch_iteration(i, other_handlers)
         if not other_handlers:
             break
         _invoke(other_handlers)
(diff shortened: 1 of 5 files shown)


johnsca:
I like this a lot and it seems like it would be very helpful in debugging.  It looks like the NullTracer is missing some methods (see failing tests) but otherwise it looks good.  IIRC, debug logging is off by default now, so maybe there would be no real penalty to having it always on at DEBUG level, but my recollection should be verified.

stub42:
Ta. I'll put some lipstick on this. I was thinking that maybe some of the existing logging could move to the tracer, such as which handler is being run, which would make the logging pluggable, in one place to help maintain visual consistency, and allow the tracer to maintain state such as which handler is running and which handlers have been run (perhaps for an end-of-hook report of the complete flow). But I'll stick with the basics first :)

stub42:
Done cleanup and tests. I needed to adjust the API so cope with handlers that don't need to run again disappear from the active handlers list when the main loop is iterating.

PTAL.

stub42:
The 'stable' travis job passed. The 'edge' channel failed due to the migration of charm-tools to classic confinement, and the travis job needs a tweak.

PR #195, opened 2018-10-02
Fix Python 3.4 support and export get_flags

stub42:
charms.reactive could not be imported under Python 3.4, and charms being built with charms.reactive 1.1.0 and 1.1.1 will fail under Trusty.

This also exports flags.get_flags to the charms.reactive namespace. It's needed by both the snap and apt layers, and done here to make the PPython 3.4 fix easier.


johnsca:
Travis failures are just Travis issues.

PR #199, opened 2018-10-04
Update docs around hook.template and call out removing apt package

johnsca:


docs/layer-basic.md
```diff
@@ -39,20 +39,30 @@ decorators of the [charms.reactive][] library:
...
+A layer can implement other hooks (e.g., `metrics`) by putting them in the `hooks`
 directory.
 
 ```eval_rst
 .. note:: Because ``update-status`` is invoked every 5 minutes, you should take
    care to ensure that your reactive handlers only invoke expensive operations
    when absolutely necessary.  It is recommended that you use helpers like
-   :func:`@when_file_changed <charms.reactive.decorators.when_file_changed>`,
-   and :func:`@data_changed <charms.reactive.helpers.data_changed>` to ensure
+   :func:`@data_changed <charms.reactive.helpers.data_changed>` to ensure
```
stub42: data_changed is a decorator?

PR #213, opened 2019-07-10
Add comment about flags executing every hook

hyperbolic2346:
I tripped up on this and thought it would be good to point this out in the docs. I'm not sure if this is the best place for it and I'm surely open to rewording. I just was here reading and cory_fu had to remind me about it while trying to figure out some code.

### DIFF
--- a/docs/managed-flags.rst
+++ b/docs/managed-flags.rst
@@ -15,6 +15,15 @@ directly; any base layer should instead use
 :func:`~charms.reactive.flags.register_trigger` to "wrap" the automatic flag
 with a layer-specific flag that can be safely used within that layer.
 
+It is worth mentioning that any set flag will execute on every hook run as
+long as it stays set. This means that code inside a function wrapped in,
+for example, ``endpoint.{endpoint_name}.joined`` will run every time hooks
+run when .joined is true. This could be due to something completely unrelated
+such as relation data changing or even an update-status hook. It is best
+not to rely on such things due to the obfusificative nature of the code as
+it makes it harder for someone to reason about, but it is a good point to
+remember.
+
 The flags that are set by the framework are:
 
 +----------------------------------------------+------------------------------------------------------------+


docs/managed-flags.rst
```diff
@@ -15,6 +15,15 @@ directly; any base layer should instead use
 :func:`~charms.reactive.flags.register_trigger` to "wrap" the automatic flag
 with a layer-specific flag that can be safely used within that layer.
 
+It is worth mentioning that any set flag will execute on every hook run as
+long as it stays set. This means that code inside a function wrapped in,
+for example, ``endpoint.{endpoint_name}.joined`` will run every time hooks
```
dshcherb: I would say that "all handlers will be evaluated against a set of flags currently set. A new evaluation will happen after every handler execution within the same Juju hook execution".

"<flag_name> will run" sounds a bit confusing.

johnsca:
In addition to dshcherb's wording improvement, this seems a bit out of place on that page, since it documents what flags are managed by the framework and not how behavior is triggered based on flags.  It would fit better in the [Discovery and Dispatch](https://charmsreactive.readthedocs.io/en/latest/dispatch.html) section, but that's under Internals and Advanced so probably isn't as visible.  I should also note that this behavior is explicitly called out in the documentation for every individual decorator (e.g., [`when_all`](https://charmsreactive.readthedocs.io/en/latest/charms.reactive.decorators.html#charms.reactive.decorators.when_all)), though that's arguably even less visible.

Perhaps we need a top-level section before Automatic Flags that gives a higher-level summary of how handlers work and when they're triggered.

hyperbolic2346:
@johnsca I'm totally not again moving this somewhere. I debated even submitting the PR, but I thought if I got caught by this that others might as well. I just know it isn't visible enough yet for me to truly grok it. I'm just trying to raise awareness to it, but I'm not sure the best way to do that.

johnsca:
@hyperbolic2346 Yeah, I'm definitely +1 to increasing the visibility, but I also struggle with the best way to accomplish it.

PR #224, opened 2020-07-15
Exclude .pyc and __pycache__ files from discovery

johnsca:
Fixes https://bugs.launchpad.net/juju/+bug/1887696


ktsakalozos:
LGTM +1

charms/reactive/bus.py
```diff
@@ -419,7 +419,11 @@ def discover():
     for search_dir in ('reactive', 'hooks/reactive', 'hooks/relations'):
         search_path = os.path.join(hookenv.charm_dir(), search_dir)
         for dirpath, dirnames, filenames in os.walk(search_path):
+            if os.path.basename(dirpath) == '__pycache__':
+                continue
             for filename in filenames:
+                if filename.endswith('.pyc'):
```
knkski, line 425: Do we need to worry about the other types of compiled Python files, like `*.pyo`?

Would it make sense to just allow files that either don't have an extensions, or have a `.sh` extension?

johnsca: Yeah, that seems reasonable.   TBH, I highly doubt anyone is actually using the ExternalHandler API at all, so making it more restrictive is unlikely to cause any issues.

charms/reactive/bus.py
```diff
@@ -443,12 +445,16 @@ def _load_module(root, filepath):
 
 
 def _register_handlers_from_file(root, filepath):
+    exec_whitelist = ('', '.py', '.sh')
     no_exec_blacklist = (
-        '.md', '.yaml', '.txt', '.ini',
         'makefile', '.gitignore',
-        'copyright', 'license')
+        'copyright', 'license',)
```
knkski, line 451: Autoformatting change? Seems unnecessary to just add a comma

charms/reactive/bus.py
```diff
@@ -443,12 +445,16 @@ def _load_module(root, filepath):
 
 
 def _register_handlers_from_file(root, filepath):
+    exec_whitelist = ('', '.py', '.sh')
     no_exec_blacklist = (
```
knkski, line 449: Instead of a blacklist, would it make sense to check that the first line of the script is some sort of shebang? If it's not, then the shell won't run it anyways.

I see below that you're using `#!/bin/false` in non-scripts, but TBH, I would expect to be able to define a valid handler with that shebang.

johnsca: Technically, the handlers could be any executable, not just a shell script.

charms/reactive/bus.py
johnsca: Copy-pasta.  I'll remove.

charms/reactive/bus.py
knkski: Check for a shebang or `+x`?

johnsca: https://github.com/juju-solutions/charms.reactive/blob/master/charms/reactive/bus.py#L455

knkski: if we're already checking for `+x`, then we shouldn't need a blacklist, since nobodies[1] going to mark `copyright` as executable, right?

[1] It might happen accidentally, hence why checking for the shebang, but I think the more important check is `+x`

knkski: In other words, the logic would be:

 - If `*.py`, then it's a reactive file
 - If it's `+x` and binary, then it's a binary handler
 - If it's `+x` and has a shebang, then it's a script handler
 - If it's `+x` and no shebang, then it's an erroneously-marked `+x` file that should be ignored
 - If it's not `+x` or `*.py`, then it can be ignored

johnsca: Unfortunately, there was an issue (#54) with random files getting marked as executable (I think due to Windows development) which is why the blacklist was added.  While not having the explicit blacklist would be nice, I don't think deciding if a file is binary or not is trivial and TBH, I don't think it's worth putting any more effort into this code since it's pretty deprecated already.  I think the whitelist addition is sufficient to exclude most files that would accidentally get picked up and the blacklist is good enough for the remaining common cases.

cjohnston1158:
application-ambassador: 19:26:00 DEBUG unit.ambassador/0.install PermissionError: [Errno 13] Permission denied: '/var/lib/juju/agents/unit-ambassador-0/charm/hooks/relations/ambassador/CODEOWNERS'

PR #225, opened 2020-07-16
Drop blacklist in favor of shebang and binary check

johnsca:
Some of the K8s charms are still getting errors after #224 due to files not in the blacklist, so it looks like we need to do the shebang and binary check that @knkski suggested after all.

### DIFF
--- a/charms/reactive/bus.py
+++ b/charms/reactive/bus.py
@@ -444,19 +444,19 @@ def _load_module(root, filepath):
     return importlib.import_module(package + module)
 
 
+def _is_external_handler(filepath):
+    if not os.access(filepath, os.X_OK):
+        return False
+    with open(filepath, 'rb') as fp:
+        bytes = fp.read(1024)
+    if bytes.startswith(b'#!'):
+        return True
+    textchars = bytearray({7, 8, 9, 10, 12, 13, 27} | set(range(0x20, 0x100)) - {0x7f})
+    return bool(bytes.translate(None, textchars))
+
+
 def _register_handlers_from_file(root, filepath):
-    exec_whitelist = ('', '.py', '.sh')
-    no_exec_blacklist = (
-        'makefile', '.gitignore',
-        'copyright', 'license')
-    _, ext = os.path.splitext(filepath)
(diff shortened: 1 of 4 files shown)


charms/reactive/bus.py
```diff
@@ -444,19 +444,19 @@ def _load_module(root, filepath):
...
 
 
+def _is_external_handler(filepath):
+    if not os.access(filepath, os.X_OK):
+        return False
+    with open(filepath, 'rb') as fp:
+        bytes = fp.read(1024)
+    if bytes.startswith(b'#!'):
+        return True
+    textchars = bytearray({7, 8, 9, 10, 12, 13, 27} | set(range(0x20, 0x100)) - {0x7f})
```
knkski, line 454: Can this get a comment as to what exactly this check is for / why these numbers in particular? I could understand an ASCII check of `0x00-0x7F`, but not sure about these. If you're checking for a particular codepage, would it make sense to do something like this?

```python
try:
    bytes.decode('utf-8')
    return True
except UnicodeDecodeError:
    return False
```

johnsca: Yeah, though binary files should probably be considered as executables, but I also included the file extension whitelist to account for things like executable tarballs.

charms/reactive/bus.py
```diff
@@ -444,19 +444,25 @@ def _load_module(root, filepath):
...
+    if not os.access(filepath, os.X_OK):
+        return False
     _, ext = os.path.splitext(filepath)
-    if ext not in exec_whitelist:
-        # Don't load handlers unless they match the whitelist
-        return
-    if filepath.lower().endswith(no_exec_blacklist):
-        # Don't load handlers which match the blacklist
-        return
+    if ext not in ('', '.sh'):
```
knkski, line 451: Definitely an edge case, but do we need to worry about `*.bash`, `*.fish`, `*.zsh`, etc?

johnsca: Given my track record so far, probably.  ;)  But my preference is to be too restrictive by default because, as I said before, I consider this a deprecated and mostly unused feature anyway.

PR #227, opened 2020-12-04
Add relation factory discovery and trigger callbacks

johnsca:
Allows additional relation factories to be registered by installed libraries via entry points. This allows for alternative relation API implementations (other than `RelationBase` and `Endpoint`) to be provided by external libraries, including ones which can be implemented by libraries rather than (the rather limited) interface layers.

This also adds the ability register flag triggers which invoke callbacks rather than simply setting or clearing another flag. Of course, this won't work via the CLI or in bash reactive charms, but no one should be using those these days anyway.

These features are to enable a "shim" [library](https://github.com/juju-solutions/ops-reactive-interface/pull/1) to allow interface API libraries targeted at the new operator framework to work with legacy reactive charms that for whatever reason can't be converted immediately. This should enable us to start converting interface layers to the new framework without having to maintain duplicate implementations.


joedborg (approved):

johnsca:
For clarification, this is intended for an upcoming feature and due to the lax pinning in [`layer:basic`](https://github.com/juju-solutions/layer-basic/blob/master/wheelhouse.txt#L11) and the fact that Charmed Kubernetes is currently in a code-freeze for the 1.20 release next week, this should **not** be released to PyPI until after that, at least (even though I'm personally fairly confident it shouldn't have any impact on existing charms).  I still really need to find time to review https://github.com/juju/charm-tools/pull/585 so that this can become less of an issue going forward.

johnsca:
> even though I'm personally fairly confident it shouldn't have any impact on existing charms

I just love saying things like this *just before* Travis finally gets around to running and I realize I didn't test against Python 3.5 and `iter_entry_points` has a different calling convention than `entry_points` (which I somehow missed in all of the examples I looked at online).

charms/reactive/flags.py
```diff
@@ -83,6 +85,8 @@ def set_flag(flag, value=None):
             set_flag(flag_name)
         for flag_name in trigger['clear_flag']:
             clear_flag(flag_name)
+        for callback in trigger['callbacks']:
+            callback()
```
ajkavanagh, line 89: I'm slightly nervous about the callback being triggered within the handling of the current handler, rather than being scheduled as the next handler to run?  But, to be fair, I've not fully understood how the callbacks are supposed to interact with handlers in terms of migrated interfaces to op-framework code?

charms/reactive/flags.py
```diff
@@ -141,31 +147,34 @@ def toggle_flag(flag, should_set):
...
 
     Note: Flag triggers are handled at the same time that the given flag is set.
 
     :param str when: Flag to trigger on when it is set.
     :param str when_not: Flag to trigger on when it is cleared.
-    :param str set_flag: If given, this flag will be set when `when` is set.
-    :param str clear_flag: If given, this flag will be cleared when `when` is set.
+    :param str set_flag: If given, this flag will be set when the relevant flag is changed.
+    :param str clear_flag: If given, this flag will be cleared when the relevant flag is changed.
+    :param str callback: If given, this callback will be invoked when the relevant flag is changed.
```
ajkavanagh, line 160: is callback a `str`?  It think it's and `Optional[Callable[]]` or something like that?

ajkavanagh (commented):
I think it looks great.  I've just got a query (not a blocker) re: *when* the callbacks should be called in relation to set_flag and clear_flag calls.  i.e. what's the "principal or least surprise" when thinking in terms of ops-framework, or thinking in 'charms.reactive' framework?

charms/reactive/flags.py
johnsca: The callback isn't intended to be a handler, and should really be limited to a minimal operation akin to setting or clearing a flag.

The specific use-case I had in mind for this is an `is_changed` property which is mapped to the `endpoint.{relation_name}.changed` flag, where clearing the flag would then automatically invoke `inst.is_changed = False` which would presumably update a hash value of relevant data in stored state so that the change is considered "handled." This is [handled automatically by the library using this feature](https://github.com/juju-solutions/ops-reactive-interface/pull/1/files#diff-be35484ccd89c06a5d619f4c4cbfc0aa1c541b7152f79f255b75178c566e12c5R121-R126) so that the interface APIs can just keep all charms.reactive-related code in a single `manage_flags` method (since that is guaranteed to only be invoked when used in a reactive charm) and otherwise assume that changes will be acknowledged via the property. Then, reactive charms can interact with the flag in the natural "reactive" fashion: have a handler triggered by the flag, handle the change, and clear the flag to indicate that it's done, without having to worry about forgetting to clear the property as well.

However, I'm thinking that this may be a bit too much magic behavior and maybe it would be safer in the end to just put a bit more onus on the reactive developer by saying that the `.changed` flag will be set automatically, but it's up to the charm to clear it and if the specific interface API also requires a property to be cleared, you'll have to remember to do that as well.

charms/reactive/flags.py
johnsca: Right.  Just expected to be a `lambda` expression or similar.  See example in above reply.

charms/reactive/flags.py
ajkavanagh: I think, based on this discussion, that the callback should be kept.  If a pattern can be established for migration of interfaces to ops-framework with interoperability back to charms.reactive code, then a little bit of magic is a fine thing :)  So I'm broadly in favour of that and the code.

ajkavanagh (approved):
I think, based on the comments, that this is ready to go.

PR #229, opened 2021-01-04
Switch to GitHub Workflows for PR tests

johnsca:
Travis is getting worse and worse, so this switches to GitHub Workflows for PR tests, and also simplifies the integration test to use the simple test charm already in the repo rather than the external Ghost charm.


joedborg (approved):
LGTM!

PR #232, opened 2021-02-19
Supporting app level relation data

xtrusia:
I need some advice to implement this.

Please review my code if it is proper or there is good way

### DIFF
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -447,7 +447,7 @@ class Relation:
         return self._data
 
     @property
-    def to_publish_app:
+    def to_publish_app(self):
         """
         Added separate method as initially We can't recognize this is app level
         relation data or unit relation data.
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -446,6 +446,20 @@ class Relation:
         return self._data
 
     @property
+    def to_publish_app
+        """
+        Added separate method as initially We can't recognize this is app level
+        relation data or unit relation data.
+        """
+        if self._data is None:
+            # using JSONUnitDataView though it's name includes unit.
+            self._data = JSONUnitDataView(
+                hookenv.relation_get(app=self.application_name,
+                                     rid=self.relation_id),
+                writeable=True)
+        return self._data
+
+    @property
     def to_publish_raw(self):
         """
         This is the raw relation data that the local unit publishes so it is
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -502,17 +502,6 @@ class Relation:
         """
         return self.to_publish.data
 
-    @property
-    def app(self):
-        return self._app
-
-    @app.setter
-    def app(self, app):
-        if app is not True:
-            self._app = False
-        else:
-            self._app = True
-
     def _flush_data(self):
         """
         If this relation's local unit data has been modified, publish it on the
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -468,10 +468,40 @@ class Relation:
         if self._app_data is None:
             # using JSONUnitDataView though it's name includes unit.
             self._app_data = JSONUnitDataView(
-                hookenv.relation_get(app=self.application_name,
+                hookenv.relation_get(app=hookenv.application_name(),
                                      rid=self.relation_id),
                 writeable=True)
         return self._app_data
+
+    @property
+    def to_publish_raw(self):
+        """
+        This is the raw relation data that the local unit publishes so it is
+        visible to all related units. It is a writeable
+        :class:`~charms.reactive.endpoints.UnitDataView`. **Only use this
+        for backwards compatibility with interfaces that do not use JSON
+        encoding.** Use
+        :attr:`~charms.reactive.endpoints.Relation.to_publish` instead.
+
+        Changes to this data are published at the end of a succesfull hook. The
+        data is reset when a hook fails.
+        """
+        return self.to_publish.data
+
+    @property
+    def to_publish_app_raw(self):
+        """
+        This is the raw relation data that the app publishes so it is
+        visible to all related units. It is a writeable (by the leader only)
+        :class:`~charms.reactive.endpoints.UnitDataView`. **Only use this
+        for backwards compatibility with interfaces that do not use JSON
+        encoding.** Use
+        :attr:`~charms.reactive.endpoints.Relation.to_publish` instead.
+
+        Changes to this data are published at the end of a succesfull hook. The
+        data is reset when a hook fails.
+        """
+        return self.to_publish_app.data
 
     @property
     def received_app(self):
@@ -488,19 +518,12 @@ class Relation:
         return self._remote_app_data
 
     @property
-    def to_publish_raw(self):
+    def received_app_raw(self):
         """
-        This is the raw relation data that the local unit publishes so it is
-        visible to all related units. It is a writeable
-        :class:`~charms.reactive.endpoints.UnitDataView`. **Only use this
-        for backwards compatibility with interfaces that do not use JSON
-        encoding.** Use
-        :attr:`~charms.reactive.endpoints.Relation.to_publish` instead.
-
-        Changes to this data are published at the end of a succesfull hook. The
-        data is reset when a hook fails.
+        A :class:`~charms.reactive.endpoints.UnitDataView` of the raw app-level
+        data received from this remote unit over the relation.
         """
-        return self.to_publish.data
+        return self.received_app.raw_data
 
     def _flush_data(self):
         """
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -341,6 +341,7 @@ class Relation:
         self._units = None
         self._departed_units = None
         self._data = None
+        # to check relation has application relation data
         self._app = False
 
     @property
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -341,8 +341,8 @@ class Relation:
         self._units = None
         self._departed_units = None
         self._data = None
-        # to check relation has application relation data
-        self._app = False
+        self._app_data = None
+        self._remote_app_data = None
 
     @property
     def relation_id(self):
--- a/charms/reactive/endpoints.py
+++ b/charms/reactive/endpoints.py
@@ -447,7 +447,7 @@ class Relation:
         return self._data
 
     @property
-    def to_publish_app
+    def to_publish_app:
         """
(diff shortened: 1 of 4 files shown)


charms/reactive/endpoints.py
```diff
@@ -341,6 +341,8 @@ def __init__(self, relation_id):
         self._units = None
         self._departed_units = None
         self._data = None
+        # to check relation has application relation data
+        self._app = False
```
johnsca, line 345: Instead of marking the entire relation as app vs unit, it would better to just have separate data buckets. There are actually potential use-cases for using both app and unit level data on the relation. This is also missing a way to read remote app data.

```suggestion
        self._app_data = None
        self._remote_app_data = None
```

charms/reactive/endpoints.py
```diff
@@ -444,6 +446,20 @@ def to_publish(self):
...
+        Added separate method as initially We can't recognize this is app level
+        relation data or unit relation data.
+        """
+        if self._data is None:
+            # using JSONUnitDataView though it's name includes unit.
+            self._data = JSONUnitDataView(
+                hookenv.relation_get(app=self.application_name,
+                                     rid=self.relation_id),
+                writeable=True)
+        return self._data
```
johnsca, line 461: +1 to separate property, but as above, it should use a second data bucket. And we also need a way to read the remote app data.

```suggestion
    @property
    def to_publish_app(self):
        """
        This is the relation data that the local app publishes so it is
        visible to all related units. Use this to communicate with related
        apps. It is a writeable
        :class:`~charms.reactive.endpoints.JSONUnitDataView`.

        Only the leader can set the app-level relation data.

        All values stored in this collection will be automatically JSON
        encoded when they are published. This means that they need to be JSON
        serializable! Mappings stored in this collection will be encoded with
        sorted keys, to ensure that the encoded representation will only change
        if the actual data changes.

        Changes to this data are published at the end of a succesfull hook. The
        data is reset when a hook fails.
        """
        if self._app_data is None:
            # using JSONUnitDataView though it's name includes unit.
            self._app_data = JSONUnitDataView(
                hookenv.relation_get(app=self.application_name,
                                     rid=self.relation_id),
                writeable=True)
        return self._app_data

    @property
    def received_app(self):
        """
        A :class:`~charms.reactive.endpoints.JSONUnitDataView` of the app-level
        data received from this remote unit over the relation, with values
        being automatically decoded as JSON.
        """
        if self._remote_app_data is None:
            # using JSONUnitDataView though it's name includes unit.
            self._remote_app_data = JSONUnitDataView(hookenv.relation_get(
                app=self.application_name,
                rid=self.relation.relation_id))
        return self._remote_app_data
```

charms/reactive/endpoints.py
```diff
@@ -459,13 +475,24 @@ def to_publish_raw(self):
...
+    def app(self):
+        return self._app
+
+    @app.setter
+    def app(self, app):
+        if app is not True:
+            self._app = False
+        else:
+            self._app = True
+
```
johnsca, line 488:
```suggestion
```

charms/reactive/endpoints.py
```diff
@@ -459,13 +475,24 @@ def to_publish_raw(self):
...
     def _flush_data(self):
         """
         If this relation's local unit data has been modified, publish it on the
         relation. This should be automatically called.
         """
         if self._data and self._data.modified:
-            hookenv.relation_set(self.relation_id, dict(self.to_publish.data))
+            hookenv.relation_set(self.relation_id, dict(self.to_publish.data), app=self._app)
```
johnsca, line 495:
```suggestion
            hookenv.relation_set(self.relation_id, dict(self.to_publish.data))
        if self._app_data and self._app_data.modified:
            hookenv.relation_set(self.relation_id, dict(self.to_publish_app.data), app=True)
```

charms/reactive/patterns/request_response.py
```diff
@@ -128,6 +128,20 @@ def _load(cls, request_sources):
                 request.response = cls.RESPONSE_CLASS._load(request)
                 cls._cache[request.request_id] = request
 
+    @classmethod
+    def _load_app(cls, relations):
+        for source in relations:
+            if source.application_name is None:
+                continue
+            app_requests = source.to_publish_app
```
johnsca, line 136: This is going to fail with a `permission denied` error from Juju on non-leader units.  I ran into the same issue with the newer `loadbalancer` interface, and [worked around it](https://github.com/juju-solutions/loadbalancer-interface/pull/13) by skipping reading the requests and reading the responses directly, but I'm not entirely sure how that would look in this implementation.  I guess `RequesterEndpoint.responses` would need to load the (app-level) responses directly rather than trying to load the (app-level) requests first.

charms/reactive/patterns/request_response.py
```diff
@@ -128,6 +128,20 @@ def _load(cls, request_sources):
...
+            for key, app_request_data in app_requests.items():
+                if not key.startswith('request_'):
+                    continue
+                source.app = True
+                request = cls(source, source['request_id'])
+                request.response = cls.RESPONSE_CLASS._load(request)
```
johnsca, line 142: The response class is also going to need a `_load_app` method which will need to read the `received_app` collection suggested above.

johnsca (changes requested):
Comments and suggestions inline.  Mostly this is just missing a way to read the app level data from the remote side~~, but there is an issue with the way the request / response pattern is currently implemented that doesn't quite work with how Juju handles app level data (the permission denied issue)~~.

**Edit:** The permission denied issue would be a problem for the general case, but the `prometheus-manual` interface is the only interface layer that I know of using that set of classes and for that case, non-leaders don't need to see the response so it's not a concern.  Just ensure that the charm wraps any usage of this in an is-leader check and it should be fine.

charms/reactive/patterns/request_response.py
johnsca: Thinking about it more, while this is a problem in general, it shouldn't be a concern for the `prometheus-manual` interface, since only the leader should care if the request failed or not (for the `loadbalancer` interface, the non-leaders still need to see the returned address, but that's not the case for Prometheus).

charms/reactive/patterns/request_response.py
xtrusia: > The response class is also going to need a `_load_app` method which will need to read the `received_app` collection suggested above.

BaseResponse's _load just returns response do i need to make _load_app inside BaseReponse class? and how can i call received_app in this _load_app if I make it?

`   
    @classmethod
    def _load(cls, request):
        response = cls(request)
        if response._key not in response._source_data:
            return None  # no response found
        else:
            return response
`

charms/reactive/patterns/request_response.py
xtrusia: In my case I wasn't able to see any permission denied error on non leader unit when leader-change and manual-job is triggered.

charms/reactive/patterns/request_response.py
johnsca: The `_load` method on both `BaseRequest` and `BaseResponse` return instances of the class from the data source.  The question is where that data comes from, and there are four possibilities each for both `RequestBase` and `ResponseBase`:

  * `unit.received` for data coming from a single unit on the remote side of the relation
  * `relation.received_app` for data coming from the leader of the remote side of the unit to represent the entire app (new in this PR)
  * `unit.to_publish` for data that this unit published for itself only
  * `relation.to_publish_app` for data that the leader of this app published for this entire app (new in this PR, and can only be read or written by the leader)

The original set of classes was written assuming that the data would only ever come in over unit-level data (since app-level data wasn't available then). It also was (poorly, by me) designed to try to track whether the data was being read or published (hence the `is_received` and `_source_data` attributes), so needs to handle both cases.  (Doing it again, the classes would just be de/serializers for a given data bag, and the relation class would be responsible for keeping track of which data bag a given request or response needs to be associated with, since it's much easier to do cleanly at that level.)

All that said, the issue is that you've added a way to read request data from the app-level relation data, but you haven't added a way to write request data to the app-level relation data nor have you added a way to either read or write response data from/to app-level relation data.

johnsca (commented):
Apologies for the long lag time between reviews on this. This library is really just in maintenance mode now that it's deprecated in favor of the Operator Framework.  I also strongly believe that there are better ways to solve [the specific bug](https://bugs.launchpad.net/charm-prometheus2/+bug/1899706) that this is for, but either way I created a card for myself to track this so I shouldn't lose track of it again.

johnsca:
I was thinking about this more, and since I'm 95% certain that the Request / Response pattern from the library is only used in the one interface implementation, maybe it would make sense to just move that part of the code out to live directly in the interface library and pare this PR down to just adding the app-level relation data support to the base `Endpoint` class.  That would make it significantly easier to land here and rev on the parts that the interface needs in that repo.

xtrusia:
> I was thinking about this more, and since I'm 95% certain that the Request / Response pattern from the library is only used in the one interface implementation, maybe it would make sense to just move that part of the code out to live directly in the interface library and pare this PR down to just adding the app-level relation data support to the base `Endpoint` class. That would make it significantly easier to land here and rev on the parts that the interface needs separately in that repo.

Hello

I think you meant I only need to submit for base Endpoint from previous commit and move request response changes to interface_prometheus_manual . so I removed code related to request_response. if there is anything I misunderstood, please correct me.

Thanks a lot

johnsca (approved):
Given the streamlining of the PR and my fixes, I think this is GTG.

charms/reactive/relations.py
```diff
@@ -474,6 +474,7 @@ def set_state(self, state, scope=None):
         See :meth:`conversation` and :meth:`Conversation.set_state`.
         """
         self.conversation(scope).set_state(state)
+    set_flag = set_state
```
ajkavanagh, line 477: The formatting looks really weird on this file; is this mixed tabs and spaces?

charms/reactive/relations.py
```diff
@@ -748,6 +754,7 @@ def is_state(self, state):
         if not value:
             return False
         return self.key in value['conversations']
+    is_flag_set = is_state
```
ajkavanagh, line 757: This is unreachable code?

charms/reactive/relations.py
johnsca: Ah, I missed that these are still in here.  They're actually method aliases, but they should have been cleaned up with the rest of the streamlining of the PR.

johnsca: Oops, these changes come from #233.  I'm not sure why they're showing up in this PR.

johnsca: Rebased to resolve.

ajkavanagh (approved):
LGTM

johnsca:
Released in 1.5.0

PR #234, opened 2021-04-09
Fix and test explicit req ID

johnsca:
From [lp:1899706][] it came up that passing an explicit `request_id` to the request / response pattern wasn't de-duping across units of a relation the way that it should. This adds a test for that and fixes an issue with a spurious `request_id can't be modified` error that I encountered.

[lp:1899706]: https://bugs.launchpad.net/charm-kubernetes-master/+bug/1899706


ajkavanagh (approved):
LGTM

PR #235, opened 2022-03-11
Fixing wrong reference

xtrusia:
Fixing wrong reference

relation_id is in Endpoint class, but self.relation. returns relation.

so it should be self.relation_id instead of self.relation.relation_id

Actually I think this haven't used before, as LP[1] is the cause of this implementation. ( prior patch was : https://github.com/juju-solutions/charms.reactive/pull/232 )

Im trying to make patch for app relation data. but prior patch has bug. 
This bug is not public I think but when I tries to use prior patch, I face issue.

[1]
LP : https://bugs.launchpad.net/charm-kubernetes-master/+bug/1899706


ajkavanagh:
> Fixing wrong reference

Please could you supply a bit more context to the PR?  e.g. what bug is it fixing?  There should also be a unit test to verify the change (particularly if there was no coverage of the code previously).

xtrusia:
> > Fixing wrong reference
> 
> Please could you supply a bit more context to the PR? e.g. what bug is it fixing? There should also be a unit test to verify the change (particularly if there was no coverage of the code previously).

xtrusia:
accidently i clicked closed button.. sorry

.github/workflows/tests.yaml
```diff
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.5, 3.6, 3.7, 3.8, 3.9]
+        python: [3.6, 3.7, 3.8, 3.9]
```
ajkavanagh, line 9: Technically, this would be a different PR; essentially, you're removing support for py35 from the library which would be mixed in with this commit.

ajkavanagh (commented):
Still would like to see a bug for this issue, please? e.g. Is this code silently breaking other charms that are using the same function `to_publish_raw`?  What charm is it failing on?  Thanks.

.github/workflows/tests.yaml
xtrusia: I removed this because github action ( testing ) failed.

xtrusia: thanks and i updated top comment

ajkavanagh (changes requested):
Please could you remove the *.swo files from the repository.  In general, if your editor generates these, you can place `*.swo` in your global `.gitignore` and they will never be included in commits.  Thanks.

xtrusia:
thanks, I removed temp file and re-added py 3.5 for testing.

testing for 3.5 will fail but didn't remove it as you said it is needed for supporting xenial

Thanks.

.github/workflows/tests.yaml
ajkavanagh: > I removed this because github action ( testing ) failed.

Okay, I now see *why* the github action failed.  i.e. because github themselves have removed the py35 options.  We could work around that, but there's probably no point now.  It's fine to remove 3.5 from the github workflow actions.  Please could you explain a bit more *why* you are removing things? e.g. "I removed this because the github action failed because github have removed py35 support from their default workflows."

ajkavanagh:
Okay, so re: verification, it would be good to see a charm that uses this?  e.g. link to charm code that will use this (and probably what you are testing with?)  It's just *really* hard to see how this works without seeing it in use.  Thanks!

xtrusia:
> Okay, so re: verification, it would be good to see a charm that uses this? e.g. link to charm code that will use this (and probably what you are testing with?) It's just _really_ hard to see how this works without seeing it in use. Thanks!

hey, it is new feature so please refer to this
this code should be merged for fixing my case.
I think there is no other charms using this.

https://github.com/juju-solutions/interface-prometheus-manual/pull/6/files

ajkavanagh:
> > Okay, so re: verification, it would be good to see a charm that uses this? e.g. link to charm code that will use this (and probably what you are testing with?) It's just _really_ hard to see how this works without seeing it in use. Thanks!
> 
> hey, it is new feature so please refer to this this code should be merged for fixing my case. I think there is no other charms using this.
> 
> https://github.com/juju-solutions/interface-prometheus-manual/pull/6/files

Okay, let's finish this off; if you strip out the py3.5 github workflow test (as it's no longer supported, as you originally discovered), then we'll merge this.  Thanks.

Issue #238, opened 2022-05-04
Docs:  invalid type param on `register_trigger`


addyess:
https://github.com/juju-solutions/charms.reactive/blob/8496d04e6497200ca8530d84eff4a3191509ed0f/charms/reactive/flags.py#L161

Invalid doc-string states `register_trigger(..., callback= ...)` is a `str` argument, when in fact it should be a method which takes no

PR #239, opened 2022-05-04
Update doc strings for `register_trigger(...)`

addyess:
Addresses issue https://github.com/juju-solutions/charms.reactive/issues/238

### DIFF
--- a/charms/reactive/flags.py
+++ b/charms/reactive/flags.py
@@ -158,7 +158,8 @@ def register_trigger(when=None, when_not=None, set_flag=None, clear_flag=None, c
     :param str when_not: Flag to trigger on when it is cleared.
     :param str set_flag: If given, this flag will be set when the relevant flag is changed.
     :param str clear_flag: If given, this flag will be cleared when the relevant flag is changed.
-    :param Callable[[], Any] callback: If given, this callback will be invoked when the relevant flag is changed.
+    :param Optional[Callable[[], Any]] callback: 
+                       If given, this callback will be invoked when the relevant flag is changed.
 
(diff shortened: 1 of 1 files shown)


charms/reactive/flags.py
```diff
@@ -158,7 +158,7 @@ def register_trigger(when=None, when_not=None, set_flag=None, clear_flag=None, c
     :param str when_not: Flag to trigger on when it is cleared.
     :param str set_flag: If given, this flag will be set when the relevant flag is changed.
     :param str clear_flag: If given, this flag will be cleared when the relevant flag is changed.
-    :param str callback: If given, this callback will be invoked when the relevant flag is changed.
+    :param Callable[[], Any] callback: If given, this callback will be invoked when the relevant flag is changed.
```
ajkavanagh, line 161: If I'm being a bit picky, I think this should be `Optional[Callable[[], Any]]` as the parameter has a `None` default.  Obviously, that would require updating the other params as well, as they all seem to be `Optional`.  Thoughts?

addyess: we can be as PICKY as we like.  I'd happily update them all.  Note these are just the doc strings and i didn't go full blown into importing `typing`.

If you think it's wise, we can do that too.

ajkavanagh: I would love to switch charms.reactive to typing.  Problem is that it still supports py36 (and maybe even py35 charms) so it's awkward to add it in (hence why we are using `:type ...` in other libraries).  It may be time to make the break, but then we should version the library properly at that point.  I think we need a **plan** for it, especially as these are now *legacy* charms.

kwmonroe (approved):

PR #242, opened 2022-12-14
Fix AttributeError when accessing received_app

rodrigogansobarbieri:
When received_app property is accessed but the
UnitDataView's data is None, the method tries
to get the value for the key requested, but
throws AttributeError. It is semantically better
to avoid access and return the default value for
the higher layers so this scenario does not have
to be specifically handled there.


rodrigogansobarbieri:
for context https://pastebin.canonical.com/p/VPR4Q43Dfb/

javacruft:
Rebased under PR: #246

PR #243, opened 2022-12-16
Fix application_name = None when single cluster relation

rodrigogansobarbieri:
When a cluster relation only has one member, the
self.application_name property used when reading
application databag returns None during a window
of opportunity that is repeated every time a new
hook runs. Subsequent handlers eventually set this property but the first handler that runs causes
issues (in vault charm for example, by failing to get the current certs and then assuming they don't
exist, therefore recreating certs every time
update-status runs).


javacruft:
Github actions and tests are now fixed in the master branch - please rebase to pick these up.

That said this commit introduces a test regression  - please look into this to resolve

tests/test_endpoints.py
```diff
@@ -408,9 +403,9 @@ def test_receive_app(self):
         Endpoint._startup()
         tep = Endpoint.from_name('test-endpoint')
 
-        self.assertEqual(tep.relations[0].received_app_raw, {'simple': 'value', 'complex': '[1, 2]'})
+        self.assertEqual(tep.relations[0].received_app_raw, {'app-key': 'value'})
         self.assertEqual(tep.relations[1].received_app_raw, {})
-        self.assertEqual(tep.relations[0].received_app, {'simple': 'value', 'complex': [1, 2]})
+        self.assertEqual(tep.relations[0].received_app, {'app-key': 'value'})
```
javacruft, line 408: The above removes part of the test data set - specifically the more complex component?

tests/test_endpoints.py
```diff
@@ -74,18 +74,14 @@ def setUp(self):
         self.relations = {
             'test-endpoint': [
                 {
-                    'local': {'app-key': 'value'},
-                    'local/0': {'key': 'value'},
-                    'unit': {'simple': 'value', 'complex': '[1, 2]'},
```
javacruft, line 79: This is where the complex component of the app data bag is removed - this seems somewhat adjunct to the renaming of the units to match the remote application name?

javacruft:
OK so I spent some time looking at the original code in the test case changed by this PR to understand what the original intent of the unit and application naming was.

The use of "local" and "local/0" relates to the local application.

The use of "unit" and "unit/X" relates to the remote application.

I'm concerned that by flattening this all into 'myapp' we've lost an element of the unit testing.

There is also some removal of complex types from the application data base data sets used for testing.

Obviously for a peer relation the remote and local units all belong to the same application so a test case for this type of interaction will need to be different to a provide/requires typed test which might be why you had a challenge with the original test failures.

charms/reactive/endpoints.py
```diff
@@ -513,7 +513,7 @@ def received_app(self):
         if self._remote_app_data is None:
             # using JSONUnitDataView though it's name includes unit.
             self._remote_app_data = JSONUnitDataView(hookenv.relation_get(
-                app=self.application_name,
+                app=hookenv.application_name(),
```
javacruft, line 516: OK so this semantically changes the behaviour of this code which is why the original unit tests fail.

hookenv.application_name() always returns the local application name.

self.application_name in this context relates to the remote application name - so this change would work OK for a peer relation but not for a standard provides or requires relation

javacruft:
```
    @property
    def application_name(self):
        """
        The name of the remote application for this relation, or ``None``.

        This is equivalent to::
        
            relation.units[0].unit_name.split('/')[0]
        """
        if self._application_name is None and self.units:
            self._application_name = self.units[0].unit_name.split('/')[0]
        return self._application_name
```
for reference

javacruft:
It feels like the Endpoint code does not do a great job with a peer relation type - I'm assuming the issue you hit originally was in this part of the codebase for the vault-ha interface:

  https://github.com/rodrigogansobarbieri/charm-interface-vault-peer/blob/main/peers.py#L60

either we need a more general change to enable better comprehension of a peer interface in reactive OR you could just use "to_publish_app_raw" in your interface code and leverage the nuance that for a peer relation, the local app data bag == the remote app data bag

rodrigogansobarbieri:
Thanks for the review James! I hope to have resolved your concerns with the latest patchset I force-pushed. I restored the original unit tests and added a new test class only for the corner case. Please re-review when you can.

charms/reactive/endpoints.py
```diff
@@ -377,8 +377,12 @@ def application_name(self):
...
             relation.units[0].unit_name.split('/')[0]
         """
-        if self._application_name is None and self.units:
-            self._application_name = self.units[0].unit_name.split('/')[0]
+        if self._application_name is None:
+            if self.units:
+                self._application_name = self.units[0].unit_name.split('/')[0]
+            else:
+                # NOTE(ganso): This is a single unit peer relation corner case
+                return hookenv.application_name()
```
javacruft, line 385: is there any harm in assigning this to self._application_name to avoid repeated lookups during a single hook execution?

javacruft (commented):
Thanks for the revise review with the standalone test case for this specific issue.

One comment but generally LGTM .

charms/reactive/endpoints.py
rodrigogansobarbieri: hey James. I had left it as it is because right now it is addressing a problem that occurs only when there is a single cluster unit, which is rarely seen on production deployments, so the performance impact is not that significant. My concern is that we are introducing a change in the library that is used by every charm, but right now we are testing just a single charm with this updated code, we don't fully know the impact of the change on other charms or other corner cases we don't know about. What I suspect could be a negative impact of assigning the value to the attribute is that if there is a window of opportunity that happens in other charms or other corner cases we don't know about (such as maybe a relation-joined hook running concurrently with something else), with the original code and proposed fix, that problem could possibly correct itself as part of the hook execution, as the attribute would continue to be None and re-evaluate next time the code is reached (which it does on almost every reactive handle). If we assign the value, then it will not be re-evaluated anymore. On the other hand, if it causes a problem it will allow us to spot and fix them (hopefully not after it is running in production). Overall, IMO, the risk/benefit isn't worth it, but I'm ok if you prefer to have it. I have made the change and tested in my lab and it works fine, but only tested this specific scenario.

